PR #6, opened 2019-05-16
Development

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds ktlint
Adds detekt (static code analysis)
Adds opendistro-job-scheduler as extended plugin along with basic boilerplate to use Job Scheduler


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementPlugin.kt
```diff
@@ -15,10 +15,55 @@
...
 import org.elasticsearch.plugins.Plugin
 
-internal class IndexStateManagementPlugin: Plugin() {
+internal class IndexStateManagementPlugin : JobSchedulerExtension, Plugin() {
+
+    private val logger = LogManager.getLogger(javaClass)
+
     companion object {
         const val PLUGIN_NAME = "opendistro-ism"
+        const val POLICY_BASE_URI = "/_opendistro/_ism/policies"
```
zengyan-amazon: do we have a convention about the REST uri? such as when to use "_"?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/ManagedIndex.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParser.Token
+import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+import java.io.IOException
+import java.time.Instant
+
+// TODO: This is an unfinished ManagedIndex data class; just needed for the jobParser
+
+data class ManagedIndex(
```
zengyan-amazon: shall we name it as something like `ManagedIndexConfig`?  "Index" has too many meaning, it sounds a little confusing to me.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementPlugin.kt
dbbaughe: Yes, following conventions from [Alerting](https://github.com/opendistro-for-elasticsearch/alerting/blob/56c49d9fe8c8332be88a9295298ded8ae5d35172/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/AlertingPlugin.kt#L83)

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/ManagedIndex.kt
dbbaughe: I prefer `ManagedIndex` over `ManagedIndexConfig`; and it follows convention of the other job we currently have in Alerting: `Monitor` instead of `MonitorConfig`.

PR #7, opened 2019-05-16
Adds opendistro-ism-config mappings

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds opendistro-ism-config index mappings


src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -0,0 +1,321 @@
...
+                  "timeout": {
+                    "type": "text"
+                  },
+                  "retry": {
+                    "properties": {
+                      "count": {
+                        "type": "integer"
+                      },
+                      "backoff": {
+                        "type": "keyword"
```
mihirsoni: What does `backoff` means, form text it looks like it should be boolean ?

dbbaughe: It's the type of backoff, such as exponential or linear. So it would look like this:

```
{
    // ... other action specific settings
    "timeout": "1h",
    "retry": {
        "count": 3,
        "backoff": "exponential",
        "delay": "5s"
    }
}
```

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -0,0 +1,321 @@
...
+            "type": "nested",
+            "properties": {
+              "name": {
+                "type": "keyword"
+              },
+              "actions": {
+                "type": "nested",
+                "properties": {
+                  "timeout": {
+                    "type": "text"
```
mihirsoni: I think we can either convert them to `keyword` or tell ES to skip analyzing.

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -0,0 +1,321 @@
...
+                  "start_time": {
+                    "type": "date",
+                    "format": "strict_date_time||epoch_millis"
+                  }
+                }
+              },
+              "cron": {
+                "properties": {
+                  "expression": {
+                    "type": "text"
```
mihirsoni: Same as previous one.

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -0,0 +1,321 @@
...
+          },
+          "last_updated_time": {
+            "type": "date",
+            "format": "strict_date_time||epoch_millis"
+          },
+          "default_state": {
+            "type": "keyword"
+          },
+          "states": {
+            "type": "nested",
```
zengyan-amazon: shall we also use strict mapping for this nested object `"dynamic": "strict"`?

src/main/resources/mappings/opendistro-ism-config.json
dbbaughe: Changed to keyword

src/main/resources/mappings/opendistro-ism-config.json
dbbaughe: Changed to keyword

PR #8, opened 2019-05-23
Policy, update to gradle, mappings fix

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds Policy data class (WIP, missing State and DefaultNotification data classes)
Updates gradle build to allow gradle tasks without path to zip
Fixes index mappings

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.indexstatemanagement.models
+
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.instant
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.optionalTimeField
+import org.elasticsearch.common.lucene.uid.Versions
+import org.elasticsearch.common.xcontent.ToXContent
+import org.elasticsearch.common.xcontent.ToXContentObject
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParser.Token
+import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+import java.io.IOException
+import java.time.Instant
+
+data class Policy(
+    val id: String = NO_ID,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
```diff
@@ -0,0 +1,129 @@
...
+                    STATES_FIELD -> {
+                        ensureExpectedToken(Token.START_ARRAY, xcp.currentToken(), xcp::getTokenLocation)
+                        while (xcp.nextToken() != Token.END_ARRAY) {
+                            // TODO: states.add(State.parse(xcp))
+                        }
+                    }
+                }
+            }
+
+            ensureExpectedToken(Token.END_OBJECT, xcp.nextToken(), xcp::getTokenLocation)
```
qreshi: Is this considered redundant since the ```while``` exit condition is ```Token.END_OBJECT```? If not, we may want to add this to [ManagedIndex.kt](https://github.com/opendistro-for-elasticsearch/index-management/blob/development/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/ManagedIndex.kt) as well

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
```diff
@@ -0,0 +1,129 @@
...
+    val defaultState: String,
+    val states: List<Map<String, Any>> // TODO: Implement List<State>
+) : ToXContentObject {
+
+    fun toXContent(builder: XContentBuilder): XContentBuilder {
+        return toXContent(builder, ToXContent.EMPTY_PARAMS)
+    }
+
+    override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        builder
```
qreshi: I like the formatting on this, it's more readable. We should change [ManagedIndex's](https://github.com/opendistro-for-elasticsearch/index-management/blob/development/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/ManagedIndex.kt#L58) ```toXContent()``` method to keep it consistent.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
```diff
@@ -0,0 +1,129 @@
...
+                .endObject()
+            .endObject()
+        return builder
+    }
+
+    companion object {
+        const val POLICY_TYPE = "policy"
+        const val NAME_FIELD = "name"
+        const val NO_ID = ""
+        const val NO_VERSION = Versions.NOT_FOUND
```
qreshi: Is there a reason why ```NO_VERSION``` here is ```Versions.NOT_FOUND``` but ```1L``` for ```ManagedIndex```?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
dbbaughe: Good catch, meant for the parseWithType for that last END_OBJET when we're parsing { "policy": { ... } }

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
dbbaughe: Yep, will change

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Policy.kt
dbbaughe: More explicit, will be changing ManagedIndex

PR #12, opened 2019-06-03
State class, updates to gradle

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds State data class (WIP, need to add Action and Transition data classes)
Updated detekt configuration in Gradle to ignore comments (ex. "TODO:") for development purposes


detekt.yml
```diff
@@ -1,2 +1,6 @@
 autoCorrect: false
 failFast: false
+
+style:
```
dbbaughe: Can we add a comment to remove this before initial release since this is temporary while we are allowing TODO comments

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
```diff
@@ -0,0 +1,93 @@
...
+
+import org.elasticsearch.common.lucene.uid.Versions
+import org.elasticsearch.common.xcontent.ToXContent
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParser.Token
+import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+import java.io.IOException
+
+data class State(
```
dbbaughe: Why do we need id/version in State?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
```diff
@@ -0,0 +1,93 @@
...
+    val id: String = NO_ID,
+    val version: Long = NO_VERSION,
+    val stateName: String,
```
dbbaughe: Nit: Change to just "name"

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
```diff
@@ -0,0 +1,93 @@
...
+                        }
+                    }
+                }
+            }
+
+            return State(
+                id,
+                version,
+                stateName = requireNotNull(name) { "State name is null" },
+                actions = requireNotNull(actions) { "State actions are null" },
```
dbbaughe: Remove requireNotNull, you are initializing actions as a mutable list, should never be null. And change to to actions.toList()

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
```diff
@@ -0,0 +1,93 @@
...
+                transitions = requireNotNull(transitions) { "State transitions are null" }
```
dbbaughe: Remove requireNotNull, you are initializing transitions as a mutable list, should never be null. And change to to transitions.toList()

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
```diff
@@ -0,0 +1,93 @@
...
+
+data class State(
+    val id: String = NO_ID,
+    val version: Long = NO_VERSION,
+    val stateName: String,
+    val actions: List<Map<String, Any>>, // TODO: Implement List<Action>
+    val transitions: List<Map<String, Any>> // TODO: Implement List<Transition>
+) : ToXContent {
+
+    fun toXContent(builder: XContentBuilder): XContentBuilder =
```
dbbaughe: Shouldn't need this on State; if you saw this from Policy it was because of the PutPolicy API

detekt.yml
qreshi: Done.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
qreshi: Good catch, it's not needed. Removed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
qreshi: Changed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
qreshi: Done.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
qreshi: Done.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
qreshi: Removed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/State.kt
```diff
@@ -0,0 +1,82 @@
...
+            .startObject()
+                .field(NAME_FIELD, name)
+                .field(ACTIONS_FIELD, actions.toTypedArray())
+                .field(TRANSITIONS_FIELD, transitions.toTypedArray())
+            .endObject()
+        return builder
+    }
+
+    companion object {
+        const val NAME_FIELD = "state"
```
dbbaughe: I think in the mappings we are using "name" instead of "state"

qreshi: Changed.

PR #13, opened 2019-06-04
Transition class, update to tests

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds Transition data class
Fixes issue with Policy parsing test
Adds validation testing for Transition and Conditions


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -0,0 +1,96 @@
...
+import com.amazon.opendistroforelasticsearch.jobscheduler.spi.schedule.ScheduleParser
+import org.elasticsearch.common.xcontent.ToXContent
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParser.Token
+import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+import java.io.IOException
+
+data class Transition(
+    val stateName: String,
```
mihirsoni: Any reason it needs to be `stateName` instead of just `state` ?

dbbaughe: Actually I think we should keep it as stateName and change "state" to "state_name" in the mappings too (and parser). Transition.name can be confusing since to someone new it might seem like the transition's name. Prefer having "name" only when it's something like State.name (as opposed to State.stateName).

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -0,0 +1,96 @@
...
+import com.amazon.opendistroforelasticsearch.jobscheduler.spi.schedule.CronSchedule
+import com.amazon.opendistroforelasticsearch.jobscheduler.spi.schedule.ScheduleParser
+import org.elasticsearch.common.xcontent.ToXContent
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParser.Token
+import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+import java.io.IOException
+
+data class Transition(
```
dbbaughe: Let's create a Conditions data class to hold the conditions:
```
Transition
    state_name: String
    conditions: Conditions
```

```
Conditions
    index_age: String?
    doc_count: Long?
    size: String?
    cron: CronSchedule?
```

And then in the future if we support different operators and grouping of conditions we can further extend Conditions with `must`, `should`, `must_not` properties to follow Elasticsearch's convention

You'll need to update the mappings for this, perhaps disabling the conditions mapping since it will eventually be nested/dynamic based on users choice.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -0,0 +1,130 @@
...
+    val cron: CronSchedule?
+) : ToXContent {
+
+    init {
+        var conditionsSet = 0
+        if (indexAge != null) conditionsSet++
+        if (docCount != null) conditionsSet++
+        if (size != null) conditionsSet++
+        if (cron != null) conditionsSet++
+
```
dbbaughe:
```
val conditionsList = listOf(indexAge, docCount, size, cron)
require(conditionsList.filterNotNull.size < 2) { "Cannot provide more than one Transition condition" }
```

dbbaughe: You should also add validation that they entered correct values for indexAge, docCount, size, i.e. don't allow "soiOSEKSOdkfsfos" for indexAge or size even though it's a valid string.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -73,16 +75,21 @@ data class Conditions(
...
+        // Validate indexAge condition
+        try {
+            TimeValue.parseTimeValue(indexAge, "")
+        } catch (e: IllegalArgumentException) {
+            throw IllegalArgumentException("Must have a valid index age Transition condition")
+        }
 
-        require(conditionsSet < 2) { "Cannot provide more than one Transition condition" }
+        // Validate size condition
+        requireNotNull(size?.toIntOrNull()) { "Must have a valid size Transition condition" }
```
dbbaughe: This will always throw an exception when the person does not use size, wrap in if (size != null)

Let's add some tests to validate these work correctly.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -73,16 +75,21 @@ data class Conditions(
...
-        if (indexAge != null) conditionsSet++
-        if (docCount != null) conditionsSet++
-        if (size != null) conditionsSet++
-        if (cron != null) conditionsSet++
+        val conditionsList = listOf(indexAge, docCount, size, cron)
+        require(conditionsList.filterNotNull().size < 2) { "Cannot provide more than one Transition condition" }
+
+        // Validate indexAge condition
+        try {
+            TimeValue.parseTimeValue(indexAge, "")
```
dbbaughe: Assuming this is:
```
public static TimeValue parseTimeValue(String sValue, String settingName) {
        Objects.requireNonNull(settingName);
        Objects.requireNonNull(sValue);
        return parseTimeValue(sValue, null, settingName);
    }
```
Seems like this will always throw NPE when indexAge is null

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -0,0 +1,137 @@
...
+        requireNotNull(size?.toIntOrNull()) { "Must have a valid size Transition condition" }
+    }
+
+    override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        builder
+            .startObject()
+                .field(INDEX_AGE_FIELD, indexAge)
+                .field(DOC_COUNT_FIELD, docCount)
+                .field(SIZE_FIELD, size)
+                .field(CRON_FIELD, cron)
```
jinsoor-amzn: we should skip the field if the value is null. Unless we want to intentionally write null fields.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
qreshi: Changed and added tests for validation.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
qreshi: Fixed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
qreshi: Done.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/XContentTests.kt
```diff
@@ -37,12 +51,41 @@ class XContentTests : ESTestCase() {
...
+
+    fun `test conditions parsing`() {
+        val conditions = nonNullRandomConditions()
+
+        val conditionsString = conditions.toJsonString()
+        val parsedConditions = Conditions.parse(parser(conditionsString))
+        assertEquals("Round tripping Conditions doesn't work", conditions, parsedConditions)
+    }
+
+    fun `test supplying more than one transition condition fails`() {
```
dbbaughe: Let's move these (from here and below) tests somewhere else, this file is for XContentTests mostly for parsing whereas these are just validating arguments.

Also can write these as:

```
assertFailsWith(IllegalArgumentException::class, "Expected IllegalArgumentException for multiple transition conditions") {
    Conditions(indexAge = "invalidIndexAge")
}
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    companion object {
+        const val STATE_FIELD = "state_name"
+        const val CONDITIONS_FIELD = "conditions"
+
+        @JvmStatic
+        @Throws(IOException::class)
+        fun parse(xcp: XContentParser): Transition {
+            lateinit var name: String
```
dbbaughe: Let's switch from lateinit usage here to just var name: String? = null
From testing it turns out if the name is missing then when doing requireNotNull(name) it'll throw an UninitializedPropertyAccessException without the message we have which seems like worse behavior.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
```diff
@@ -0,0 +1,148 @@
...
+                TimeValue.parseTimeValue(indexAge, "")
+            } catch (e: IllegalArgumentException) {
+                throw IllegalArgumentException("Must have a valid index age Transition condition")
+            }
+        }
+
+        // Validate size condition
+        if (size != null) {
+            try {
+                ByteSizeValue.parseBytesSizeValue(size, "")
```
dbbaughe: It seems like ByteSizeValue.parseBytesSize accepts "0" and "-1" for size, do we want to allow these values?

qreshi: Yeah I noticed that these were considered special values to ```ByteSizeValue```. I can't really see a use for them. If you agree, I can include additional validation to make sure size is > 0.

dbbaughe: Agreed, sounds good

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
qreshi: Removed lateinit from all model ```parse``` methods.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/models/Transition.kt
qreshi: Done.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/XContentTests.kt
qreshi: Moved argument validation tests to another file called ```ConditionsTests```

PR #21, opened 2019-06-21
Adds abstract ActionConfig, Action, Step classes. Adds DeleteAction, …

dbbaughe:
…TransitionsAction, AttemptDeleteStep, AttemptTransitionStep. Adds TransitionsActionConfig.

*Issue #, if available:*

*Description of changes:*


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/actions/TransitionsAction.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val steps = listOf(attemptTransitionStep)
+
+    override fun getSteps(): List<Step> = steps
+
+    override fun getNextStep(managedIndexMetaData: ManagedIndexMetaData): Step {
+        return attemptTransitionStep
+    }
+
+    companion object {
+        const val type = "transitions"
```
zengyan-amazon: what is this action type for? and what is the difference between action type and action config type?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/steps/AttemptTransitionStep.kt
```diff
@@ -0,0 +1,57 @@
...
+    private val logger = LogManager.getLogger(javaClass)
+
+    override suspend fun execute() {
+        val statsRequest = IndicesStatsRequest()
+                .indices(managedIndexMetaData.index).clear().docs(true)
+        val stats: IndicesStatsResponse = client.admin().indices().suspendUntil { stats(statsRequest, it) }
+        val indexCreationDate = clusterService.state().metaData().index(managedIndexMetaData.index).creationDate
+        val numDocs = stats.primaries.docs.count
+        val indexSize = stats.primaries.docs.totalSizeInBytes
+        logger.info("""
```
zengyan-amazon: multiline log may not be a good practice

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/actions/DeleteAction.kt
```diff
@@ -0,0 +1,45 @@
...
+    config: DeleteActionConfig
+) : Action(type, managedIndexMetaData) {
+
+    private val attemptDeleteStep =
+            AttemptDeleteStep(clusterService, client, config, managedIndexMetaData)
+    private val steps = listOf(attemptDeleteStep)
+
+    override fun getSteps(): List<Step> = steps
+
+    override fun getNextStep(managedIndexMetaData: ManagedIndexMetaData): Step {
```
zengyan-amazon: shall we return null if the current step is AttempDeleteStep?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/actions/TransitionsActionConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.indexstatemanagement.model.actions
```
jinsoor-amzn: should we make `actions` to singular `action` like the `model`?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/steps/AttemptDeleteStep.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.indexstatemanagement.steps
```
jinsoor-amzn: Also singular `step`? And lets make additional package to group like `delete`, `transition`, `rollover` so it will be easier to identify them.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/steps/AttemptTransitionStep.kt
dbbaughe: Yeah this was just a placeholder log for development, this step isn't finished yet.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/actions/TransitionsAction.kt
dbbaughe: Used to populate the metadata with the type of action that is being executed.
The type in this action and the type in the action config should be equal for the same actions. Will move all types out into an enum so there's no duplication.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/actions/DeleteAction.kt
dbbaughe: The name "getNextStep" might be confusing. There should never be a step after AttemptDeleteStep because for that to succeed we would of had to delete the index successfully which should delete this job.

This method is returning the step that should be executed during the execution of runJob.

Will change to getStepToExecute

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/delete/AttemptDeleteStep.kt
```diff
@@ -0,0 +1,65 @@
...
+            stepStartTime = getStepStartTime().toEpochMilli().toString()
+            // transitionTo = null
+            // stepComplete = !failed
+            // failed = failed
+            // info = info
+        )
+    }
+
+    companion object {
+        const val name = "attempt_delete"
```
jinsoor-amzn: If we don't use this static name outside of this class, we can pass this into the constructor itself and use the `name` variable from there.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/delete/AttemptDeleteStep.kt
```diff
@@ -0,0 +1,65 @@
...
+            step = name,
+            stepStartTime = getStepStartTime().toEpochMilli().toString()
+            // transitionTo = null
+            // stepComplete = !failed
+            // failed = failed
+            // info = info
+        )
+    }
+
+    companion object {
```
jinsoor-amzn: I don't now if we have best practice around this but I feel like it is easier to read `companion object` near the top of the class because I think these variables are considered `static` in kotlin. So having them at the top makes it easier to read.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -175,3 +178,43 @@ fun getSweptManagedIndexSearchRequest(): SearchRequest {
                     )
                     .query(boolQueryBuilder))
 }
+
+enum class ActionType(val type: String) {
```
jinsoor-amzn: Any reason this is not a file on its own? If we want it to be an inner class I think it fit better in `ActionConfig` to hold all the related Action data into it.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/delete/AttemptDeleteStep.kt
dbbaughe: Current recommendation is to keep it at the bottom
https://kotlinlang.org/docs/reference/coding-conventions.html#class-layout

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
dbbaughe: Moved into ActionConfig

PR #23, opened 2019-07-01
Adds rollover action and step. Updates Action/ActionConfig. Adds roll…

dbbaughe:
…over_alias setting. Adds more ManagedIndexMetaData fields

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
@@ -42,93 +42,11 @@ data class ManagedIndexMetaData(
     val stepStartTime: Long?,
     val stepCompleted: Boolean?,
     val failed: Boolean?,
+    val policyCompleted: Boolean?,
+    val rolledOver: Boolean?,
     val info: Map<String, Any>?,
     val consumedRetries: Int?
 ) : Writeable, ToXContentFragment {
-
-    companion object {
-        const val MANAGED_INDEX_METADATA = "managed_index_metadata"
-
-        const val INDEX = "index"
-        const val INDEX_UUID = "index_uuid"
-        const val POLICY_NAME = "policy_name"
-        const val POLICY_SEQ_NO = "policy_seq_no"
-        const val POLICY_PRIMARY_TERM = "policy_primary_term"
-        const val STATE = "state"
-        const val STATE_START_TIME = "state_start_time"
-        const val TRANSITION_TO = "transition_to"
-        const val ACTION_INDEX = "action_index"
-        const val ACTION = "action"
-        const val ACTION_START_TIME = "action_start_time"
-        const val STEP = "step"
-        const val STEP_START_TIME = "step_start_time"
-        const val STEP_COMPLETED = "step_completed"
-        const val FAILED = "failed"
-        const val INFO = "info"
-        const val CONSUMED_RETRIES = "consumed_retries"
-
-        fun fromMap(map: Map<String, String?>): ManagedIndexMetaData {
-            return ManagedIndexMetaData(
-                requireNotNull(map[INDEX]) { "$INDEX is null" },
-                requireNotNull(map[INDEX_UUID]) { "$INDEX_UUID is null" },
-                requireNotNull(map[POLICY_NAME]) { "$POLICY_NAME is null" },
-                map[POLICY_SEQ_NO]?.toLong(),
-                map[POLICY_PRIMARY_TERM]?.toLong(),
-                map[STATE],
-                map[STATE_START_TIME]?.toLong(),
-                map[TRANSITION_TO],
-                map[ACTION_INDEX]?.toInt(),
-                map[ACTION],
-                map[ACTION_START_TIME]?.toLong(),
-                map[STEP],
-                map[STEP_START_TIME]?.toLong(),
-                map[STEP_COMPLETED]?.toBoolean(),
-                map[FAILED]?.toBoolean(),
-                map[INFO]?.let { XContentHelper.convertToMap(JsonXContent.jsonXContent, it, false) },
-                map[CONSUMED_RETRIES]?.toInt()
-            )
-        }
-
-        fun fromStreamInput(si: StreamInput): ManagedIndexMetaData {
-            val index: String? = si.readString()
-            val indexUuid: String? = si.readString()
-            val policyName: String? = si.readString()
-            val policySeqNo: Long? = si.readOptionalLong()
-            val policyPrimaryTerm: Long? = si.readOptionalLong()
-            val state: String? = si.readOptionalString()
-            val stateStartTime: Long? = si.readOptionalLong()
-            val transitionTo: String? = si.readOptionalString()
-            val actionIndex: Int? = si.readOptionalInt()
-            val action: String? = si.readOptionalString()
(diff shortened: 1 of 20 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -147,31 +65,47 @@ data class ManagedIndexMetaData(
...
             FAILED to failed?.toString(),
+            POLICY_COMPLETED to policyCompleted?.toString(),
+            ROLLED_OVER to rolledOver?.toString(),
             INFO to info?.let { Strings.toString(XContentFactory.jsonBuilder().map(it)) },
             CONSUMED_RETRIES to consumedRetries?.toString()
         )
     }
 
     override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        // Order matters here as we are only trying to show what is needed for the customer
```
jinsoor-amzn: I am not sure about what you mean with `Order matters here`. The order here is just how the json will look when creating the respond. I think you might have misunderstood with order of `writeTo` and `fromStreamInput`. That is where the order really matter. If we have incorrect write and read we will be assign incorrect value to a variable.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -147,31 +65,47 @@ data class ManagedIndexMetaData(
...
-        if (actionStartTime != null) builder.field(ACTION_START_TIME, actionStartTime)
-        if (step != null) builder.field(STEP, step)
-        if (stepStartTime != null) builder.field(STEP_START_TIME, stepStartTime)
-        if (stepCompleted != null) builder.field(STEP_COMPLETED, stepCompleted)
+        if (policyCompleted == true) {
+            builder.field(POLICY_COMPLETED, policyCompleted)
+            return builder
+        }
+
+        if (rolledOver == true) {
```
jinsoor-amzn: Is there a reason we don't show if `rolledOver` is `false`? shouldn't this be really checking for `step` is rollover and if it then we should always be interested in this param.

dbbaughe: Yes, good catch. I think we can go with: If rolledOver is true always show, otherwise only show rolledOver value if we're in the Rollover action. Otherwise we probably don't care about rolledOver yet as we either a) haven't gotten to Rollover action or b) don't even have Rollover in our Policy.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
dbbaughe: The order I'm talking about is the order in which we check the values. If the policyCompleted is true then we do not want to show any of the other fields to the user and can return early.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/action/TransitionsActionConfig.kt
```diff
@@ -22,7 +22,10 @@ import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.Transiti
 import org.elasticsearch.client.Client
 import org.elasticsearch.cluster.service.ClusterService
 
-data class TransitionsActionConfig(val transitions: List<Transition>) : ActionConfig(ActionType.TRANSITION, null, null) {
+const val TRANSITION_ACTION_INDEX = -1
+data class TransitionsActionConfig(
+    val transitions: List<Transition>
+) : ActionConfig(ActionType.TRANSITION, null, null, TRANSITION_ACTION_INDEX) {
```
jinsoor-amzn: np:  I think it is ok to just pass in the `index` as `-1` instead of creating this `const val` outside of the class.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    private fun getAlias(): String? {
+        val alias = clusterService.state().metaData().index(managedIndexMetaData.index).getRolloverAlias()
+
+        if (alias == null) {
+            failed = true
+            info = mapOf("message" to "There is no valid rollover_alias=$alias set on ${managedIndexMetaData.index}")
+        }
```
jinsoor-amzn: I think this if block in here can be counter intuitive. From the function name it sounds like only thing this should do is `getAlias` but it has some business logic that needs to be handled outside of this function. So when I initially saw the 

```
val alias = getAlias()

alias ?: return
```

In `execute` I was unaware this is what was happen in here. I think it makes more sense to let `getAlais` to only get `alias` and let the error handling to be done in the `execute`. Other implementation that may make sense is `getAlias` throwing an exception if the `alias` is null.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -0,0 +1,152 @@
...
+            val statsRequest = IndicesStatsRequest()
+                    .indices(managedIndexMetaData.index).clear().docs(true)
+            val statsResponse: IndicesStatsResponse = client.admin().indices().suspendUntil { stats(statsRequest, it) }
+
+            if (statsResponse.status == RestStatus.OK) {
+                return statsResponse
+            }
+
+            failed = true
+            info = mapOf(
```
jinsoor-amzn: Similar comment to `getAlias` here. I think function name shows that only thing it should do it `getIndexStats` but really it does a bit more than that which and disrupt the flow of the `execute` function. In the `execute`

```
        val statsResponse = getIndexStats()

        statsResponse ?: return
```

seems like we are not handling the null case gracefully only to find out there is some error handling logic in `getIndexStats`. Lets make `get` functions only do `get` and let the error handling logic be in the flow of the `execute`.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -0,0 +1,152 @@
...
+                "message" to "Failed to get Index stats",
```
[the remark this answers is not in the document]
dbbaughe: Just by accident/habit :), fixed

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
dbbaughe: We update the info/failed different ways depending on errors, response status, etc. and it pollutes the execute method whereas I think we should try to keep execute a high level overview of what's happening in this step. I agree the naming is confusing `getAlias`, changing to a more verbose name for now until we can eventually come up with a better solution to the metadata and execute.

PR #29, opened 2019-07-09
implement open and close action

jinsoor-amzn:
*Description of changes:*
implement open and close action


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.elasticsearch.action.support.master.AcknowledgedResponse
+import org.elasticsearch.client.Client
+import org.elasticsearch.cluster.service.ClusterService
+
+class AttemptCloseStep(
+    val clusterService: ClusterService,
+    val client: Client,
+    val config: CloseActionConfig,
+    managedIndexMetaData: ManagedIndexMetaData
+) : Step("attempt_close", managedIndexMetaData) {
```
qreshi: Looks like in the last PR, Drew changed the ```name``` to be within a ```companion object```. Let's change this and ```AttemptOpenStep``` to be consistent with that.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        // Need to wait two cycles.
+        // change the start time so the job will trigger in 2 seconds.
+        updateManagedIndexConfigStartTime(managedIndexConfig, Instant.now().minusSeconds(58).toEpochMilli())
+
+        Thread.sleep(3000)
+
+        assertEquals("close", getIndexState(indexName))
+    }
+}
```
qreshi: NP: Please add a newline at the end of here and ```OpenActionIT```

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.action.CloseActionConfig
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.randomDefaultNotification
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.Locale
+
+class CloseActionIT : IndexStateManagementRestTestCase() {
+
+    fun `test basic`() {
+        val indexName = "${javaClass.simpleName.toLowerCase(Locale.getDefault())}_index"
```
qreshi: Let's change this to ```toLowerCase(Locale.ROOT)```, I believe ```Locale.getDefault()``` is called implicitly when no arguments are supplied and could lead to a format issue depending on region.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    override suspend fun execute() {
+        logger.info("Executing close on ${managedIndexMetaData.index}")
+        val closeIndexRequest = CloseIndexRequest()
+            .indices(managedIndexMetaData.index)
+
+        val response: AcknowledgedResponse = client.admin().indices().suspendUntil { close(closeIndexRequest, it) }
+        if (!response.isAcknowledged) {
+            failed = true
+            info = mapOf("message" to "Failed to close index: ${managedIndexMetaData.index}")
```
zengyan-amazon: shall we also keep the error message from the response? if any

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/action/CloseActionConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+    val index: Int
+) : ToXContentObject, ActionConfig(ActionType.CLOSE, timeout, retry, index) {
+
+    override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        builder.startObject().startObject(ActionType.CLOSE.type)
+        super.toXContent(builder, params)
+        return builder.endObject().endObject()
+    }
+
+    override fun isFragment(): Boolean = super<ToXContentObject>.isFragment()
```
qreshi: Is there a reason you preferred calling the ```super``` implementation rather than explicitly specifying ```false```? I assumed we defined this method for readability but since ```ToXContentObject```'s implementation returns ```false```, defining this and calling ```super``` would be redundant.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/open/AttemptOpenStep.kt
```diff
@@ -0,0 +1,48 @@
...
+        logger.info("Executing open on ${managedIndexMetaData.index}")
+        val openIndexRequest = OpenIndexRequest()
+            .indices(managedIndexMetaData.index)
+
+        val response: OpenIndexResponse = client.admin().indices().suspendUntil { open(openIndexRequest, it) }
+        if (!response.isAcknowledged) {
+            failed = true
+            info = mapOf("message" to "Failed to open index: ${managedIndexMetaData.index}")
+        }
+        response.isShardsAcknowledged
```
qreshi: What is this used for? I noticed the other step implementations don't have it.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/action/CloseActionConfig.kt
qreshi: Resolving this since Yan made a similar comment.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
dbbaughe: I did that because I was consuming that name somewhere else, whereas we are only using this name here so don't necessarily need it outside this class. Can make the choice to either do it for all of them or only the ones that need it.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
```diff
@@ -0,0 +1,52 @@
...
+        val managedIndexConfig = getManagedIndexConfig(indexName)
+        assertNotNull("ManagedIndexConfig is null", managedIndexConfig)
+        // change the start time so the job will trigger in 2 seconds.
+        updateManagedIndexConfigStartTime(managedIndexConfig!!, Instant.now().minusSeconds(58).toEpochMilli())
+
+        Thread.sleep(3000)
+
+        // Need to wait two cycles.
+        // change the start time so the job will trigger in 2 seconds.
+        updateManagedIndexConfigStartTime(managedIndexConfig, Instant.now().minusSeconds(58).toEpochMilli())
```
dbbaughe: Nice speed up of tests :P
We should definitely look into how we can leverage job scheduler to allow faster executions for testing.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/action/CloseActionConfig.kt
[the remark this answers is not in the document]
jinsoor-amzn: Right, normally implementing`ToXContentObject` should just "work" but in this case `ActionConfig` also implements `ToXContentFragment` which also implements `isFragment`. So there is a conflict in `CloseActionConfig` (and other class that inherits both from `ActionConfig` and `ToXContentObject`) and compiler complains about which default `isFragment` to use.

This implementation is a way to indicate `CloseActionConfig` should use the default implementation of the `ToXContentObject.isFragment`

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.randomDefaultNotification
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.Locale
+
+class CloseActionIT : IndexStateManagementRestTestCase() {
+
+    fun `test basic`() {
+        val indexName = "${javaClass.simpleName.toLowerCase(Locale.getDefault())}_index"
+        val policyId = "${javaClass.simpleName.toLowerCase(Locale.getDefault())}_testPolicyName"
```
qreshi: NP: Saw that [declaration names with acronyms](https://kotlinlang.org/docs/reference/coding-conventions.html#choosing-good-names) are capitalized when they are less than three characters (ex. ```policyID```)

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/open/AttemptOpenStep.kt
jinsoor-amzn: ah I was thinking what I should do with this and was trying some things but didn't end up needing to use it. I will remove it.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
jinsoor-amzn: nice. will do

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
jinsoor-amzn: The AcknowledgedResponse does not have any other messages so unfortunately there is nothing else to get message from :(

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
jinsoor-amzn: right so since we already have `name` as part of `Step`, I think it is redundant to create another companion object to store the name. I think we should extract the `name` as we really needed them.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
jinsoor-amzn: I wonder if we can just find a way to directly call the `JobRunner.runJob` (in our case `ManagedIndexRunner.runjob`) that should be really helpful. But for now this will do..

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
qreshi: Okay sure, let's extract ```name``` only as needed.

PR #30, opened 2019-07-12
refactor ManagedIndexMetaData into logical groups

jinsoor-amzn:
*Description of changes:*
refactor ManagedIndexMetaData into logical groups
Rename PolicyName to PolicyID


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -220,26 +222,26 @@ object ManagedIndexRunner : ScheduledJobRunner,
...
             rolledOver = false,
-            info = if (policy == null) mapOf("message" to "Could not load policy: ${managedIndexConfig.policyName}") else null,
-            consumedRetries = null
+            transitionTo = null,
+            stateMetaData = stateMetaData,
+            actionMetaData = null,
+            stepMetaData = null,
+            // TODO fix retryInfo when we implement retry logic.
+            retryInfo = RetryInfoMetaData(failed = policy == null, consumedRetries = 0),
+            info = if (policy == null) mapOf("message" to "Could not load policy: ${managedIndexConfig.policyID}") else null
```
dbbaughe: If there is a policy, we could add a message that we successfully initialized or loaded policy

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ChangePolicy.kt
```diff
@@ -53,14 +52,14 @@ data class ChangePolicy(
                 xcp.nextToken()
 
                 when (fieldName) {
-                    POLICY_NAME_FIELD -> policyName = xcp.text()
-                    STATE_FIELD -> state = xcp.textOrNull()
+                    ManagedIndexConfig.POLICY_ID_FIELD -> policyID = xcp.text()
+                    StateMetaData.STATE -> state = xcp.textOrNull()
```
dbbaughe: np: Let's add back this classes own POLICY_ID_FIELD and STATE_FIELD, I know they are technically the same, but if we for whatever reason decided to rename StateMetaData.STATE to "something_else" it shouldn't impact ChangePolicy since this one would still be "state".

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
```diff
@@ -26,11 +26,9 @@ abstract class Step(val name: String, val managedIndexMetaData: ManagedIndexMeta
     abstract fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData
 
     fun getStepStartTime(): Instant {
-        val startTimeMillis = managedIndexMetaData.stepStartTime
-        if (managedIndexMetaData.step != name || startTimeMillis == null) {
+        if (managedIndexMetaData.stepMetaData == null) {
```
dbbaughe: Do we still need a check for the step name here?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
```diff
@@ -35,11 +37,10 @@ class AttemptCloseStep(
 
     override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
         return currentMetaData.copy(
-            step = name,
-            stepStartTime = getStepStartTime().toEpochMilli(),
-            transitionTo = null,
```
dbbaughe: Do we not need to reset transitionTo to null anymore?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/delete/AttemptDeleteStep.kt
```diff
@@ -61,13 +63,10 @@ class AttemptDeleteStep(
 
     override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
         return currentMetaData.copy(
-            step = name,
             // TODO only update stepStartTime when first try of step and not retries
-            // Will update when retry logic is added
-            stepStartTime = getStepStartTime().toEpochMilli(),
-            transitionTo = null,
```
dbbaughe: Same comment about transitionTo not being reset

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/open/AttemptOpenStep.kt
```diff
@@ -35,11 +37,10 @@ class AttemptOpenStep(
 
     override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
         return currentMetaData.copy(
-            step = name,
-            stepStartTime = getStepStartTime().toEpochMilli(),
-            transitionTo = null,
```
dbbaughe: Also here

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/readonly/SetReadOnlyStep.kt
```diff
@@ -58,18 +60,17 @@ class SetReadOnlyStep(
...
+            if (errorMessage != null) mutableInfo["cause"] = errorMessage
             info = mutableInfo.toMap()
         }
     }
 
     override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
         return currentMetaData.copy(
-            step = name,
-            stepStartTime = getStepStartTime().toEpochMilli(),
-            transitionTo = null,
```
dbbaughe: Also here

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/readwrite/SetReadWriteStep.kt
```diff
@@ -58,18 +60,17 @@ class SetReadWriteStep(
...
+            if (errorMessage != null) mutableInfo["cause"] = errorMessage
             info = mutableInfo.toMap()
         }
     }
 
     override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
         return currentMetaData.copy(
-            step = name,
-            stepStartTime = getStepStartTime().toEpochMilli(),
-            transitionTo = null,
```
dbbaughe: Also here (if we don't need to reset this anymore then these are all just spam now 😄 )

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -141,12 +143,10 @@ class AttemptRolloverStep(
     // TODO: retries, stepStartTime not resetting when same step
     override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
         return currentMetaData.copy(
-            step = name,
-            stepStartTime = getStepStartTime().toEpochMilli(),
-            transitionTo = null,
```
dbbaughe: Also here

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -263,51 +266,51 @@ fun State.getActionToExecute(
...
+                actionConfig = this.actions.getOrNull(managedIndexMetaData.actionMetaData.index + 1) ?: TransitionsActionConfig(this.transitions)
             }
         }
     }
 
     return actionConfig.toAction(clusterService, client, managedIndexMetaData)
 }
 
-fun State.getUpdatedManagedIndexMetaData(managedIndexMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
+fun State.getUpdatedManagedIndexMetaData(managedIndexMetaData: ManagedIndexMetaData): StateMetaData {
```
dbbaughe: np: rename to getUpdatedStateMetaData if it's returning that now

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -263,51 +266,51 @@ fun State.getActionToExecute(
...
+        else -> stateMetaData
+    }
 }
 
-fun Action.getUpdatedManagedIndexMetaData(managedIndexMetaData: ManagedIndexMetaData, state: State): ManagedIndexMetaData {
-    val actionStartTime = if (managedIndexMetaData.state != state.name || this.config.actionIndex != managedIndexMetaData.actionIndex) {
-        Instant.now().toEpochMilli()
-    } else {
-        managedIndexMetaData.actionStartTime
+fun Action.getUpdatedManagedIndexMetaData(managedIndexMetaData: ManagedIndexMetaData, state: State): ActionMetaData {
```
dbbaughe: Same as above

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/TestHelpers.kt
```diff
@@ -54,14 +54,14 @@ import java.time.ZoneId
 import java.time.temporal.ChronoUnit
 
 fun randomPolicy(
-    name: String = ESRestTestCase.randomAlphaOfLength(10),
```
dbbaughe: Should this be description instead of id?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ChangePolicy.kt
jinsoor-amzn: Shouldn't the `ChangePolicy` follow what ever the change we make to state if it is a different name? `ChangePolicy` should be combining existing model names and add it as part of its own model since the nature of this class is to indicate what changes will happen to other models.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
jinsoor-amzn: I think changing `transistionTo` in Step is illogical since `Steps` shouldn't not need to know about `transistionTo` but since that is not the case right now I will add back in. We should refactor this to be outside of the step.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/delete/AttemptDeleteStep.kt
jinsoor-amzn: I think changing `transistionTo` in Step is illogical since `Steps` shouldn't not need to know about `transistionTo` but since that is not the case right now I will add back in. We should refactor this to be outside of the step.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/TestHelpers.kt
jinsoor-amzn: No, this was being used as Id before it Id is correct.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
jinsoor-amzn: no, StepMetaData is a data class that requires non-null Step name and StartTime. So checking if StepMetaData is not null is enough indication that the step is on first execution and can return `Instant.now()`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ChangePolicy.kt
jinsoor-amzn: There is a resource mapping the need to be mapped with certain string. So it make sense `ChangePolicy` to have its own string const that is the same as the Index mapping. I will make the changed necessary

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
jinsoor-amzn: We need to check if the step name is the same or else we need to return new start time.

I will make the change.

PR #31, opened 2019-07-17
Adds Add Policy API and Remove Policy API

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds Add Policy API
Adds Remove Policy API
Minor refactors and code reformatting
Disables TooGenericExceptionCaught in detekt while developing


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
```diff
@@ -0,0 +1,177 @@
...
+        private val indicesToAddPolicyTo: MutableList<Index> = mutableListOf()
+
+        override fun processResponse(clusterStateResponse: ClusterStateResponse) {
+            val state = clusterStateResponse.state
+            populateLists(state)
+
+            val builder = channel.newBuilder().startObject()
+            if (indicesToAddPolicyTo.isNotEmpty()) {
+                val updateSettingsRequest = UpdateSettingsRequest()
+                    .indices(*indicesToAddPolicyTo.map { it.name }.toTypedArray())
```
qreshi: Is there an alternative to using a spread operator in these cases? Given how it works internally, this could be very inefficient when an index pattern is given matching a large number of indices

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementRestTestCase.kt
```diff
@@ -208,4 +217,55 @@ abstract class IndexStateManagementRestTestCase : ESRestTestCase() {
...
+        return indexSettings[indexName]!!["settings"]!!["index.uuid"] as String
+    }
+
+    /**
+     * Compares responses returned by APIs such as those defined in [RetryFailedManagedIndexAction] and [RestAddPolicyAction]
+     *
+     * Example response with no failures:
+     * {
+     *   "failures": false,
+     *   "updated_indices": 3
```
dbbaughe: If there are no failures, let's return "failed_indices": [] still

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementRestTestCase.kt
```diff
@@ -208,4 +217,55 @@ abstract class IndexStateManagementRestTestCase : ESRestTestCase() {
...
+     *     {
+     *       "index_name": "indexName",
+     *       "index_uuid": "s1PvTKzaThWoeA43eTPYxQ"
+     *       "reason": "Reason for failure"
+     *     }
+     *   ]
+     * }
+     */
+    @Suppress("UNCHECKED_CAST")
+    protected fun assertApiResponseIsEqual(expected: Map<String, Any>, actual: Map<String, Any>) {
```
dbbaughe: "assertApiResponseIsEqual" seems like we are just comparing any generic responses, but this is specific to ones that return in the structure outlined in the comment. Either rename or make this generic.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/RestHandlerUtils.kt
```diff
@@ -30,3 +31,26 @@ const val REFRESH = "refresh"
...
+const val FAILED_INDICES = "failed_indices"
+const val UPDATED_INDICES = "updated_indices"
+
+fun buildInvalidIndexResponse(builder: XContentBuilder, failedIndices: MutableList<FailedIndex>) {
+    if (failedIndices.isNotEmpty()) {
+        builder.field(FAILURES, true)
+        builder.startArray(FAILED_INDICES)
+        for (failedIndex in failedIndices) {
+            builder.startObject()
+            builder.field("index_name", failedIndex.name)
```
dbbaughe: This could go in FailedIndex so it's reusable

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/RestHandlerUtils.kt
```diff
@@ -30,3 +31,26 @@ const val REFRESH = "refresh"
 
 const val WITH_TYPE = "with_type"
 val XCONTENT_WITHOUT_TYPE = ToXContent.MapParams(mapOf(WITH_TYPE to "false"))
+
+const val FAILURES = "failures"
+const val FAILED_INDICES = "failed_indices"
+const val UPDATED_INDICES = "updated_indices"
+
+fun buildInvalidIndexResponse(builder: XContentBuilder, failedIndices: MutableList<FailedIndex>) {
```
dbbaughe: Let's always return "failed_indices": [] (empty or with FailedIndices).

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
```diff
@@ -95,55 +114,64 @@ class RestRetryFailedManagedIndexAction(
                                     builder.field(UPDATED_INDICES, listOfIndexMetaData.size)
                                 } else {
                                     failedIndices.addAll(listOfIndexMetaData.map {
-                                        FailedIndex(it.first.name, it.first.uuid, "failed to update IndexMetaData")
```
jinsoor-amzn: np: can you check how long your max line length is? I think it looks ok for this to be a one liner and doesn't need to be broken down into multiple lines.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
```diff
@@ -95,55 +114,64 @@ class RestRetryFailedManagedIndexAction(
...
         private fun populateList(state: ClusterState, startState: String?) {
-            for (indexMetadataEntry in state.metaData.indices) {
-                val indexMetaData = indexMetadataEntry.value
+            for (indexMetaDataEntry in state.metaData.indices) {
+                val indexMetaData = indexMetaDataEntry.value
                 val managedIndexMetaData = indexMetaData.getManagedIndexMetaData()
                 when {
                     indexMetaData.getPolicyName() == null ->
-                        failedIndices.add(FailedIndex(indexMetaData.index.name, indexMetaData.index.uuid, "This index is not being managed."))
+                        failedIndices.add(
```
jinsoor-amzn: np: same here, for me it looks like better in single line. Was this part of auto formatting from intellij?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
```diff
@@ -0,0 +1,177 @@
...
+            throw IllegalArgumentException("Missing indices")
+        }
+
+        val body = if (request.hasContent()) {
+            XContentHelper.convertToMap(request.requiredContent(), false, request.xContentType).v2()
+        } else {
+            mapOf()
+        }
+
+        if (!body.containsKey("policy_id") || body["policy_id"] == null) {
```
dbbaughe: Could do something like:

val policyId = requireNotNull(body.getOrDefault("policy_id", null), "Missing policy_id")

so that you don't have to use body["policy_id"] in multiple places

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
```diff
@@ -0,0 +1,177 @@
...
+        controller.registerHandler(POST, "$ADD_POLICY_BASE_URI/{index}", this)
+    }
+
+    override fun getName(): String = "add_policy_action"
+
+    @Throws(IOException::class)
+    override fun prepareRequest(request: RestRequest, client: NodeClient): RestChannelConsumer {
+        val indices: Array<String>? = Strings.splitStringByCommaToArray(request.param("index"))
+
+        if (indices == null || indices.isEmpty()) {
```
dbbaughe: isNullOrEmpty

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
jinsoor-amzn: I also tried to find other ways to pass `vararg` variable from Kotlin to Java without using `spread operator` but couldn't find anything. I don't know if there is other way.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementRestTestCase.kt
qreshi: Looked into how ```assertEquals``` behaves and it turns out that since this map response contains either primitive values or Strings and because ```Assert``` has a method ```internalArrayEquals``` which allows two arrays to be compared based on content, we can actually do ```assertEquals(expected, actual)``` on these two response maps.

However, this would require making sure the list of failed indices in the two response maps are in the same order. I began writing a utility method to do this for the "actual" message that we currently retrieve by doing ```response.asMap()``` but thought since this would mean the person writing the tests would need to write the "expectedMessage" response in order, that it's not very intuitive. So I will keep this assertion method and call it something more descriptive.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
```diff
@@ -0,0 +1,177 @@
...
+import org.elasticsearch.rest.RestController
+import org.elasticsearch.rest.RestRequest
+import org.elasticsearch.rest.RestRequest.Method.POST
+import org.elasticsearch.rest.RestResponse
+import org.elasticsearch.rest.RestStatus
+import org.elasticsearch.rest.action.RestActionListener
+import org.elasticsearch.rest.action.RestResponseListener
+import java.io.IOException
+
+class RestAddPolicyAction(settings: Settings, controller: RestController) : BaseRestHandler(settings) {
```
dbbaughe: Have one major concern about the current implementation for AddPolicy.
We do not want to allow an Add Policy to go through if a policy already exists. But, the current implementation is doing a ClusterStateRequest to get the updated cluster state, parsing to find indices to update, and then doing an index settings update. This has the potential to overwrite another users request if they both do this around the same time.

User_1: Calls AddPolicy API on index_a
User_2: Calls AddPolicy API on index_a
The coordination nodes that accepted request will do a cluster state request. Both requests see no policy on index_a. Both will do an index settings update. One of them will override the other. Both will send OK 200 responses back to the users.

Ideally we should be doing a transport action that executes the logic of iterating through cluster state indices and updating the index settings on the master node itself. And then that transport request can return the status, failed_indices, updated_indices back to the coordinating node.

This might not be a blocker right now though, as with our current design, even if both go through, at worst it'll just override a managed index that has not been initialized yet. As it takes 1 minute before the initialization happens.

To protect us in the mean time, can you look into adding timeouts such that if this doesn't succeed within (30 seconds? to be safe) then it'll fail? Has to be from the time the cluster state request goes out to the end of the index settings update.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/RestHandlerUtils.kt
qreshi: Defined ```toXContent()``` method for FailedIndex to clean this up

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/RestHandlerUtils.kt
qreshi: Done.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
qreshi: Nice suggestion, done.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
qreshi: Fixed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestAddPolicyAction.kt
qreshi: Good catch. For the time being I added a timeout of 30 seconds like you said.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
qreshi: My mistake, I didn't notice we changed the detekt configuration to a maximum of 150 lines. However, I think we should stick with 120 max lines. In my opinion, anything more than 120 affects readability and if we're going off of Kotlin's style conventions we shouldn't really be going over that number.

I went ahead and reverted my format changes to ```FailedIndex``` calls to use 150 lines as the max for now. I'm open to discussing this further if there are strong preferences on either side.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
qreshi: Changed to single line.

jinsoor-amzn:
I have approved this PR. I will check also after resolving the merge conflict.

PR #32, opened 2019-07-22
Adds destination data classes

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds destination data classes. These are temporary imports from the Alerting plugin. These should not be long lived in this package as we should be pulling the notification subproject out of alerting and make it a standalone consumable.


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/destination/Chime.kt
```diff
@@ -0,0 +1,80 @@
...
+ * A value object that represents a Chime message. Chime message will be
+ * submitted to the Chime destination
+ *
+ * Temporary import from alerting, this will be removed once we pull notifications out of
+ * alerting so all plugins can consume and use.
+ */
+data class Chime(val url: String) : ToXContent {
+
+    init {
+        require(!Strings.isNullOrEmpty(url)) { "URL is null or empty" }
```
jinsoor-amzn: will this throw correct exception message if the `url` is null? I think it would throw a generic exception from Kotlin giving exception message about how we can assign `null` to `String`. Lets use convention we are using in ISM code where we check before we pass into the constructor.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/destination/Slack.kt
```diff
@@ -0,0 +1,80 @@
...
+ * A value object that represents a Slack message. Slack message will be
+ * submitted to the Slack destination
+ *
+ * Temporary import from alerting, this will be removed once we pull notifications out of
+ * alerting so all plugins can consume and use.
+ */
+data class Slack(val url: String) : ToXContent {
+
+    init {
+        require(!Strings.isNullOrEmpty(url)) { "URL is null or empty" }
```
jinsoor-amzn: Same here. Lets check required before we pass into the constructor.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/destination/Chime.kt
dbbaughe: This was copied from Alerting (temporary fix until Notifications is made standalone). I changed url to be nullable so that we can throw the specific message here.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/DestinationTests.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    fun `test custom webhook destination with host and no url`() {
+        try {
+            val customWebhook = CustomWebhook(null, null, "abc.com", 80, null, emptyMap(), emptyMap(), null, null)
+            assertEquals("host is manipulated", customWebhook.host, "abc.com")
+        } catch (ignored: IllegalArgumentException) {
+        }
+    }
+
+    fun `test custom webhook destination with url and host`() {
```
qreshi: This test says ```with url and host``` but only a ```url``` was given

dbbaughe: This is just a copy from Alerting, it shouldn't live in this repo for long and will be removed so should be fine for now.

PR #33, opened 2019-07-23
catch generic exception in all steps

jinsoor-amzn:
*Description of changes:*
1. catch generic exception in all steps
1. Add retry and consumedRetries to ActionMetaData

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
@@ -19,7 +19,7 @@ import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateMana
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.getManagedIndexMetaData
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.getPolicyID
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.ManagedIndexMetaData
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedindexmetadata.RetryInfoMetaData
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedindexmetadata.PolicyRetryInfoMetaData
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.transport.action.updateindexmetadata.UpdateManagedIndexMetaDataAction
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.transport.action.updateindexmetadata.UpdateManagedIndexMetaDataRequest
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.FailedIndex
@@ -63,6 +63,7 @@ class RestRetryFailedManagedIndexAction(
         return "retry_failed_managed_index"
     }
 
+    @Suppress("SpreadOperator")
     override fun prepareRequest(request: RestRequest, client: NodeClient): RestChannelConsumer {
         val indices: Array<String>? = Strings.splitStringByCommaToArray(request.param("index"))
         if (indices == null || indices.isEmpty()) {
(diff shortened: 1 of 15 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexAction.kt
```diff
@@ -143,15 +144,15 @@ class RestRetryFailedManagedIndexAction(
                         failedIndices.add(FailedIndex(indexMetaData.index.name, indexMetaData.index.uuid, "This index is not being managed."))
                     managedIndexMetaData == null ->
                         failedIndices.add(FailedIndex(indexMetaData.index.name, indexMetaData.index.uuid, "There is no IndexMetaData information"))
-                    managedIndexMetaData.retryInfo == null || !managedIndexMetaData.retryInfo.failed ->
+                    managedIndexMetaData.policyRetryInfo == null || !managedIndexMetaData.policyRetryInfo.failed ->
```
dbbaughe: Need to check for action retry related info too

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/transition/AttemptTransitionStep.kt
```diff
@@ -82,8 +81,6 @@ class AttemptTransitionStep(
             transitionTo = stateName,
```
dbbaughe: Need to add try/catch around the above execute logic too

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -348,13 +350,12 @@ fun ManagedIndexMetaData.getUpdatedManagedIndexMetaData(
...
         info = updatedStepMetaData.info
-
     )
 }
 
 val ManagedIndexMetaData.isSuccessfulDelete: Boolean
     get() = (this.actionMetaData != null && this.actionMetaData.name == ActionConfig.ActionType.DELETE.type) &&
             (this.stepMetaData != null && this.stepMetaData.name == AttemptDeleteStep.name && this.stepMetaData.completed) &&
-            (this.retryInfo != null && !this.retryInfo.failed)
+            (this.policyRetryInfo != null && !this.policyRetryInfo.failed)
```
dbbaughe: Need to check action retry info here?

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/CloseActionIT.kt
```diff
@@ -68,4 +68,46 @@ class CloseActionIT : IndexStateManagementRestTestCase() {
...
+            defaultState = states[0].name,
+            states = states
+        )
+        createPolicy(policy, policyID)
+        createIndex(indexName, null)
+        closeIndex(indexName)
+
+        assertEquals("close", getIndexState(indexName))
+
+        addPolicyToIndex(indexName, policyID)
```
qreshi: NP: Should add policy before closing the index initially since this creates an unrealistic workflow. Normally if someone's tries to add a policy to a closed index via the API, it will fail, so it might be confusing for that to pass here.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -356,6 +356,21 @@ fun ManagedIndexMetaData.getUpdatedManagedIndexMetaData(
 }
 
 val ManagedIndexMetaData.isSuccessfulDelete: Boolean
-    get() = (this.actionMetaData != null && this.actionMetaData.name == ActionConfig.ActionType.DELETE.type) &&
+    get() = (this.actionMetaData != null && this.actionMetaData.name == ActionConfig.ActionType.DELETE.type && !this.actionMetaData.failed) &&
             (this.stepMetaData != null && this.stepMetaData.name == AttemptDeleteStep.name && this.stepMetaData.completed) &&
             (this.policyRetryInfo != null && !this.policyRetryInfo.failed)
+
+val ManagedIndexMetaData.isFailed: Boolean
```
dbbaughe: np: This could be cleaned up. I don't think we need that middle else if. They all return so don't need else if, can just use if.
```
val ManagedIndexMetaData.isFailed: Boolean
    get() {
        // If PolicyRetryInfo is failed then the ManagedIndex has failed.
        if (this.policyRetryInfo?.failed == true) return true
        // If ActionMetaData is not null and some action is failed. Then the ManagedIndex has failed.
        if (this.actionMetaData?.failed == true) return true
        return false
    }
```

PR #51, opened 2019-07-26
Adds policy_id to policy mappings to support fuzzy searching

dbbaughe:
*Issue #, if available:*

*Description of changes:*
We need to support fuzzy searching by policy_id as we expect people to be able to use it when adding policies and searching for them. Elasticsearch doesn't support it on the _uid field so we need to add it ourself to our mappings/documents.

### DIFF
--- a/src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/IndexStateManagementRestApiIT.kt
+++ b/src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/IndexStateManagementRestApiIT.kt
@@ -206,7 +206,7 @@ class IndexStateManagementRestApiIT : IndexStateManagementRestTestCase() {
 
     @Throws(Exception::class)
     fun `test able to fuzzy search policies`() {
-        val policy = createRandomPolicy()
+        val policy = createRandomPolicy(refresh = true)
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/IndexStateManagementRestApiIT.kt
```diff
@@ -199,4 +203,26 @@ class IndexStateManagementRestApiIT : IndexStateManagementRestTestCase() {
...
+                        "query": "*${policy.id.substring(4, 7)}*"
+                    }
+                }
+            }
+        """.trimIndent()
+        val response = client().makeRequest("POST", "$INDEX_STATE_MANAGEMENT_INDEX/_search", emptyMap(),
+                StringEntity(request, APPLICATION_JSON))
+        assertEquals("Request failed", RestStatus.OK, response.restStatus())
+        val searchResponse = SearchResponse.fromXContent(createParser(jsonXContent, response.entity.content))
+        assertTrue("Did not find policy using fuzzy search", searchResponse.hits.hits.size != 1)
```
qreshi: Shouldn't we be asserting ```searchResponse.hits.hits.size == 1``` is true if it was able to find it?

PR #56, opened 2019-08-07
Adds force merge action

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds force merge action and implementation
Adds ActionProperties to ActionMetaData
**Note:** Still working on adding tests but the implementation is complete and can be reviewed


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/AttemptSetReadOnlyStep.kt
```diff
@@ -0,0 +1,124 @@
...
+            if (!indexSetToReadOnly) return
+        }
+
+        // Complete step since index is read-only
+        stepStatus = StepStatus.COMPLETED
+        info = mapOf("message" to "Set index to read-only")
+    }
+
+    private suspend fun isReadOnly(indexName: String): Boolean {
+        val getSettingsResponse: GetSettingsResponse = client.admin().indices()
```
dbbaughe: Wrap in try catch

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
```diff
@@ -0,0 +1,130 @@
...
+            if (errorMessage != null) mutableInfo["cause"] = errorMessage
+            info = mutableInfo.toMap()
+        }
+
+        return null
+    }
+
+    override fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData {
+        return currentMetaData.copy(
+            wasReadOnly = false, // Resetting this as the force_merge action is over
```
dbbaughe: remove

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/AttemptRevertReadOnlyStep.kt
```diff
@@ -0,0 +1,104 @@
...
+        // If the index was not read-only before the force merge, set back to read-write
+        if (managedIndexMetaData.wasReadOnly == false) {
+            val indexSetToReadWrite = setIndexToReadWrite(indexName)
+
+            // If setIndexToReadWrite returns false, updating settings failed and failed info was already updated. Can return early.
+            if (!indexSetToReadWrite) return
+
+            info = mapOf("message" to "Set index to read-write")
+        } else {
+            logger.info("The index [$indexName] was read-only prior to the force_merge action, no action necessary")
```
jinsoor-amzn: lower the log level to debug.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeAction.kt
```diff
@@ -0,0 +1,74 @@
...
+        val currentStepStatus = stepMetaData.stepStatus
+
+        // If the current step has completed, return the next step
+        if (currentStepStatus == Step.StepStatus.COMPLETED) {
+            return when (currentStep) {
+                AttemptSetReadOnlyStep.name -> attemptCallForceMergeStep
+                AttemptCallForceMergeStep.name -> waitForForceMergeStep
+                WaitForForceMergeStep.name -> attemptRevertReadOnlyStep
+                // Shouldn't hit this case but including it so that the when expression is exhaustive
+                else -> stepNameToStep[currentStep]!!
```
jinsoor-amzn: You can replace with `stepNameToStep[currentStep]` and check if the map returns object with non-null object. If the returned object is null then we can log the `currentStep` and throw an error.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -296,6 +296,7 @@ object ManagedIndexRunner : ScheduledJobRunner,
                 policyPrimaryTerm = policy?.primaryTerm,
                 policyCompleted = false,
                 rolledOver = false,
+                wasReadOnly = null,
```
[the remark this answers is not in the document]
qreshi: Removed ```wasReadOnly``` from being at the top ```ManagedIndexMetaData``` level and added an ```ActionProperties``` class within ```ActionMetaData``` to hold any properties we want passed along steps of a particular ```Action```.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeAction.kt
qreshi: Replaced returning ```stepNameToStep[currentStep]!!``` with throwing an ```IllegalArgumentException``` for the ```else``` case.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ActionConfigTests.kt
```diff
@@ -56,4 +57,12 @@ class ActionConfigTests : ESTestCase() {
             randomRolloverActionConfig(minDocs = 0)
         }
     }
+
+    fun `test force merge action max num segments of zero fails`() {
+        assertFailsWith(
```
dbbaughe: np: Was this over the 140 limit we currently have? this seems harder to read than
```
assertFailsWith(IllegalArgumentException::class, "Expected IllegalArgumentException for maxNumSegments less than 1") {
    randomForceMergeActionConfig(maxNumSegments = 0)
}

```

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeActionIT.kt
```diff
@@ -0,0 +1,222 @@
...
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.Locale
+
+class ForceMergeActionIT : IndexStateManagementRestTestCase() {
+
+    private val testIndexName = javaClass.simpleName.toLowerCase(Locale.ROOT)
+
+    fun `test basic workflow`() {
+        val indexName = "${testIndexName}_index"
```
dbbaughe: I had issues during testing where the cluster wasn't fully cleaned up between tests in a test class. Can you differentiate the indexName/policyId per test (just add a # or something at end)

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeActionIT.kt
```diff
@@ -0,0 +1,222 @@
...
+        createIndex(indexName, null)
+        refresh()
+
+        // Add sample data to increase segment count, passing in a delay so that multiple segments get created
+        insertSampleData(indexName, 3, 1000)
+
+        val segmentCount = getSegmentCount(indexName)
+        assertTrue("Segment count for [$indexName] was less than expected", segmentCount > 1)
+
+        // Add policy to index after data is loaded
```
dbbaughe: You should be able to just add the policyID on the initial createIndex. It takes 1 minute to initialize policy and 1 minute to execute the first action so you shouldn't have any issues with policy being run too soon.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeActionIT.kt
```diff
@@ -0,0 +1,222 @@
...
+        assertEquals("Segment count for [$indexName] after force merge is incorrect", 1, segmentCountAfterForceMerge)
+
+        // Fifth execution: Set index back to read-write since it was not originally read-only
+        updateManagedIndexConfigStartTime(managedIndexConfig, Instant.now().minusSeconds(58).toEpochMilli())
+        Thread.sleep(3000)
+
+        assertEquals("false", getIndexBlocksWriteSetting(indexName))
+
+        // Sixth execution: Transition to DeleteState
+        updateManagedIndexConfigStartTime(managedIndexConfig, Instant.now().minusSeconds(58).toEpochMilli())
```
dbbaughe: No need to test delete, we already test it in its own IT, should be fine to stop after asserting index set back to read-write

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeActionIT.kt
```diff
@@ -0,0 +1,222 @@
...
+
+    fun `test force merge on index already in read-only`() {
+        val indexName = "${testIndexName}_index"
+        val policyID = "${testIndexName}_testPolicyName"
+
+        // Create a Policy with two States, one that set index to read-only and one that force-merges
+        val readOnlyActionConfig = ReadOnlyActionConfig(index = 0)
+        val forceMergeActionConfig = ForceMergeActionConfig(maxNumSegments = 1, index = 0)
+        val states = listOf(
+            State(
```
dbbaughe: Similar to above, you could just manually set the index to read-only instead of relying on a state/action transition which is adding time to the overall testing time. We already test readonly actions in its own IT.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeActionIT.kt
```diff
@@ -0,0 +1,222 @@
...
+        refresh()
+
+        // Add sample data to increase segment count, passing in a delay so that multiple segments get created
+        insertSampleData(indexName, 3, 1000)
+
+        val segmentCount = getSegmentCount(indexName)
+        assertTrue("Segment count for [$indexName] was less than expected", segmentCount > 1)
+
+        // Add policy to index after data is loaded
+        addPolicyToIndex(indexName, policyID)
```
dbbaughe: Can just pass to createIndex instead here too

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/TestHelpers.kt
```diff
@@ -133,6 +134,15 @@ fun randomReplicaCountActionConfig(): ReplicaCountActionConfig {
     return ReplicaCountActionConfig(index = 0, numOfReplicas = ESRestTestCase.randomIntBetween(0, 200))
 }
 
+fun randomForceMergeActionConfig(
+    maxNumSegments: Int = ESRestTestCase.randomIntBetween(1, 50)
+): ForceMergeActionConfig {
+    return ForceMergeActionConfig(
```
dbbaughe: np: one line

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -402,3 +406,9 @@ fun ManagedIndexConfig.shouldChangePolicy(managedIndexMetaData: ManagedIndexMeta
 
     return true
 }
+
+val ManagedIndexMetaData.wasReadOnly: Boolean?
```
dbbaughe: Can do:
`return this.actionMetaData?.actionProperties?.getBoolean(ActionProperties.WAS_READ_ONLY) == true`
To convert the response to just a `Boolean` instead of `Boolean?`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/ActionMetaData.kt
```diff
@@ -122,6 +133,7 @@ data class ActionMetaData(
                     FAILED -> failed = xcp.booleanValue()
                     CONSUMED_RETRIES -> consumedRetries = xcp.intValue()
                     LAST_RETRY_TIME -> lastRetryTime = if (xcp.currentToken() == Token.VALUE_NULL) null else xcp.longValue()
+                    ActionProperties.ACTION_PROPERTIES -> actionProperties = ActionProperties.parse(xcp)
```
dbbaughe: You're allowing ActionProperties to be null which means this will throw an exception if it does pass null into the parse. Do a similar check that's in LAST_RETRY_TIME

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    private val logger = LogManager.getLogger(javaClass)
+    private var stepStatus = StepStatus.STARTING
+    private var info: Map<String, Any>? = null
+
+    @Suppress("TooGenericExceptionCaught") // TODO see if we can refactor to catch GenericException in Runner.
+    override suspend fun execute() {
+        val indexName = managedIndexMetaData.index
+
+        logger.info("Waiting for force merge to complete on [$indexName]")
```
dbbaughe: np: Change to "Checking if force merge is complete on [$indexName]"

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
```diff
@@ -0,0 +1,145 @@
...
+                    "Force merge still running on [$indexName] with [$shardsStillMergingSegments] shards containing unmerged segments"
+                )
+
+                stepStatus = StepStatus.CONDITION_NOT_MET
+                info = mapOf("message" to "Waiting for force merge to complete")
+            }
+        }
+    }
+
+    private fun getMaxNumSegments(): Int? {
```
dbbaughe: np: this is hard to read, something like this might be easier:
```
private fun getMaxNumSegments(): Int? {
    val actionProperties = managedIndexMetaData.actionMetaData?.actionProperties
    if (actionProperties == null) {
        stepStatus = StepStatus.FAILED
        info = mapOf("message" to "Unable to retrieve [${ActionProperties.MAX_NUM_SEGMENTS}] from ActionProperties=$actionProperties")
    }
    return actionProperties?.getInt(ActionProperties.MAX_NUM_SEGMENTS)
    }
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
```diff
@@ -0,0 +1,145 @@
...
+            val statsRequest = IndicesStatsRequest().indices(indexName)
+            val statsResponse: IndicesStatsResponse = client.admin().indices().suspendUntil { stats(statsRequest, it) }
+
+            if (statsResponse.status == RestStatus.OK) {
+                return statsResponse.shards.count { it.stats.segments.count > maxNumSegments }
+            }
+
+            stepStatus = StepStatus.FAILED
+            info = mapOf(
+                "message" to "Failed to get index stats",
```
dbbaughe: This is returning a message to the customer about an internal request we're using in a step, let's change the message to something about the force merge action instead.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
```diff
@@ -0,0 +1,145 @@
...
+                "status" to statsResponse.status,
+                "shard_failures" to statsResponse.shardFailures.map { it.toString() }
+            )
+        } catch (e: Exception) {
+            stepStatus = StepStatus.FAILED
+            val mutableInfo = mutableMapOf("message" to "Failed to get index stats")
```
dbbaughe: Same as above

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
```diff
@@ -0,0 +1,145 @@
...
+             * If there are still shards with segments merging then no action is taken and the step will be reevaluated
+             * on the next run of the ManagedIndexRunner.
+             *
+             * However, if an internal timeout of 12 hours is reached, the force_merge action will be marked as failed.
+             * This is for cases where the force merge stops silently, such as when nodes relocate during a force
+             * merge operation. Without a timeout, the segment count would stop going down as merging would no longer
+             * occur and the managed index would become stuck in this action.
+             */
+            val timeWaitingForForceMerge: Duration = Duration.between(getStepStartTime(), Instant.now())
+            if (timeWaitingForForceMerge.toSeconds() > FORCE_MERGE_TIMEOUT_IN_SECONDS) {
```
dbbaughe: What happens if we have an internal force merge timeout of 12 hours and then they set a timeout on the force merge action as 48 hours. Should we respect theirs or ours?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeAction.kt
```diff
@@ -0,0 +1,78 @@
...
+        val currentStepStatus = stepMetaData.stepStatus
+
+        // If the current step has completed, return the next step
+        if (currentStepStatus == Step.StepStatus.COMPLETED) {
+            return when (currentStep) {
+                AttemptSetReadOnlyStep.name -> attemptCallForceMergeStep
+                AttemptCallForceMergeStep.name -> waitForForceMergeStep
+                WaitForForceMergeStep.name -> attemptRevertReadOnlyStep
+                // Shouldn't hit this case but including it so that the when expression is exhaustive
+                else -> throw IllegalArgumentException("Unexpected step: [$currentStep] when getting step to execute for ForceMergeAction")
```
dbbaughe: Are you catching this somewhere? Otherwise this is throwing an uncaught exception in the middle the runner.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/ActionProperties.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.elasticsearch.common.io.stream.Writeable
+import org.elasticsearch.common.xcontent.ToXContent
+import org.elasticsearch.common.xcontent.ToXContentFragment
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParser.Token
+import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+
+/** Properties that will persist across steps of a single Action. Will be stored in the [ActionMetaData]. */
+class ActionProperties : Writeable, ToXContentFragment {
```
dbbaughe: It looks like you're throwing exceptions in here with the require/requireNotNull
Are you catching these in the places you use this or will this just kill our current runner execution?

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ActionConfigTests.kt
qreshi: These tests were written before the change to 140 character line limit. Changed the format to make it easier to read.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/ActionMetaData.kt
qreshi: Good catch, added a check.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
qreshi: Updated the messages about ```"Failed to get index stats"``` in this Step as well as ```AttemptRolloverStep``` and ```AttemptTransitionStep```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
qreshi: Agreed, it was an Intellij suggestion but did make the code harder to read.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
qreshi: Good suggestion, changed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt
qreshi: Changed this to favor ```ActionTimeout``` if given and use our internal 12 hour timeout if nothing was provided.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/ForceMergeAction.kt
qreshi: Actually reverted this to the old check for a couple reasons:
1. As you mentioned, to not possibly throw an exception in the middle of runner
2. The additional ```if``` check we have before this ```when``` statement now should prevent any cases where an unexpected ```currentStep``` is getting evaluated here. If for some reason we do reach this ```else``` it would mean that we are on the last step and the runner did not properly evaluate we were on the last step and tried to get a step to execute for some reason. If this happens, returning the currentStep (last step in this case) would be the same behavior we currently have for our other Actions

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/ActionProperties.kt
qreshi: Refactored ```ActionProperties``` to not use ```Map``` and just contain the properties as class attributes. These ```require/requireNotNull``` checks are no longer present.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -188,7 +188,7 @@ object ManagedIndexRunner : ScheduledJobRunner,
         if (updateResult && state != null && action != null && step != null && actionMetaData != null) {
             // Step null check is done in getStartingManagedIndexMetaData
             step.execute()
-            val executedManagedIndexMetaData = startingManagedIndexMetaData.getCompletedManagedIndexMetaData(state, action, step, actionMetaData)
+            val executedManagedIndexMetaData = startingManagedIndexMetaData.getCompletedManagedIndexMetaData(state, action, step)
```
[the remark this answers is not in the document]
qreshi: It cleans up the ```if``` statement a bit when we're checking if ```actionMetaData``` is null. Will leave it for now but we can remove it when we do a PR for code cleanup.

PR #59, opened 2019-08-15
Adds notification support, action, error notification, etc.

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds dependency on notification subproject from alerting
Currently we will use mavenLocal for development and expect any developers to pull down alerting and publish the notification to their local maven repository until we officially publish it to maven.
Adds notification actions, configs, steps, etc.
Adds error_notification logic
Removes need of having job-scheduler as a parent directory and relies on the job-scheduler SPI from maven
Updates mappings
Adds support for mustache templates in the notification/error_notification messages along with the context variable we pass in to be used


src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/TestHelpers.kt
```diff
@@ -163,8 +218,9 @@ fun randomChangePolicy(
     return ChangePolicy(policyID, state, emptyList())
 }
 
-fun randomDefaultNotification(): Map<String, Any>? { // TODO: DefaultNotification data class
-    return null // TODO: random DefaultNotification
+// will only return null since we dont want to send actual notifications during integ tests
+fun randomErrorNotification(): ErrorNotification? {
+    return null
```
qreshi: NP: Since Kotlin recommends [using expressions bodies for single expression functions](https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting) it may be better to do ```= null```. There are cases where we need to do this elsewhere in our code but that can be handled in upcoming code cleanups/refactors.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -65,6 +67,23 @@ data class ManagedIndexMetaData(
         )
     }
 
+    fun asTemplateArg(): Map<String, Any?> {
```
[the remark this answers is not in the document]
dbbaughe: I'd like to keep this separate from just fromMap/ToMap as this is what's specifically passed into the mustache script as a template argument (ctx variable).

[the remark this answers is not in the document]
qreshi: I see, that's fine then.

jinsoor-amzn: Can we reuse toXContent output to give back to customer so that we have same log to remove unnecessary information back to users? I know toXContent is not a Map but since it is a `XContentBuilder` should be easy to transform it back to a Map

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -44,6 +44,7 @@ data class ManagedIndexMetaData(
     val actionMetaData: ActionMetaData?,
     val stepMetaData: StepMetaData?,
     val policyRetryInfo: PolicyRetryInfoMetaData?,
+    val errorNotificationFailure: String?,
```
jinsoor-amzn: What is the reason that this error message is outside of the info? Can't we add it as part of existing info map and  add one more entry?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -441,4 +466,20 @@ object ManagedIndexRunner : ScheduledJobRunner,
             logger.error("There was an error while trying to update the policy_id ($policyID) setting for $index", e)
         }
     }
+
+    private suspend fun publishErrorNotification(policy: Policy, managedIndexMetaData: ManagedIndexMetaData) {
+        policy.errorNotification?.run {
+            errorNotificationRetryPolicy.retry(logger) {
+                withContext(Dispatchers.IO) {
+                    destination.publish(null, compileTemplate(messageTemplate, managedIndexMetaData))
```
jinsoor-amzn: Publishing message with no subject? Is it because we have more messenger type notifications?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/ActionMetaData.kt
```diff
@@ -76,6 +76,17 @@ data class ActionMetaData(
         return Strings.toString(this, false, false)
     }
 
+    fun asTemplateArg(): Map<String, Any?> {
```
jinsoor-amzn: Same comment. Lets try to reuse `toXContent` instead of creating new function which can have different logic. I already see that this is missing `ActionProperties`.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/PolicyRetryInfoMetaData.kt
```diff
@@ -47,9 +47,9 @@ data class PolicyRetryInfoMetaData(
             .field(CONSUMED_RETRIES, consumedRetries)
     }
 
-    fun getMapValueString(): String {
-        return Strings.toString(this, false, false)
-    }
+    fun getMapValueString(): String = Strings.toString(this, false, false)
+
+    fun asTemplateArg(): Map<String, Any> = mapOf(FAILED to failed, CONSUMED_RETRIES to consumedRetries)
```
jinsoor-amzn: Reuse `toXContent`.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/managedindexmetadata/StateMetaData.kt
```diff
@@ -49,9 +49,9 @@ data class StateMetaData(
             .field(START_TIME, startTime)
     }
 
-    fun getMapValueString(): String {
-        return Strings.toString(this, false, false)
-    }
+    fun getMapValueString(): String = Strings.toString(this, false, false)
+
+    fun asTemplateArg(): Map<String, Any> = mapOf(NAME to name, START_TIME to startTime)
```
jinsoor-amzn: Reuse `toXContent`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/readonly/SetReadOnlyStep.kt
```diff
@@ -57,7 +57,7 @@ class SetReadOnlyStep(
                 info = mapOf("message" to "Failed to set index to read-only")
             }
         } catch (e: Exception) {
-            stepStatus = StepStatus.STARTING
+            stepStatus = StepStatus.FAILED
```
jinsoor-amzn: good catch!

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
dbbaughe: Discussed offline and we will utilize the toXContent unless we see requests for full ManagedIndexMetaData in ctx variable.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
dbbaughe: The subject was for "SNS" destination types which we do not currently support in ISM. We might in the future and I'd like to keep the structure/calls the same as the notifications in alerting as we eventually move notifications into its own repository. This will make the integration with that easier as noting differs from alerting.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
dbbaughe: Was thinking info was mainly for step/action related info whereas errorNotificationFailure is a failure regarding their configuration of the error_notification at the root level of the policy. But, it looks like we do dump other runner/policy related info in there, so will move it inside info too.

PR #66, opened 2019-08-22
Make IT unique index names and policy names.

jinsoor-amzn:
*Description of changes:*
Make IT unique index names and policy names.


src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestRetryFailedManagedIndexActionIT.kt
```diff
@@ -34,10 +37,10 @@ class RestRetryFailedManagedIndexActionIT : IndexStateManagementRestTestCase() {
     }
 
     fun `test index list`() {
-        val indexName = "movies"
-        val indexName1 = "movies_1"
-        val indexName2 = "movies_2"
-        val indexName3 = "some_other_test"
+        val indexName = "${testIndexName}_movies"
```
qreshi: NP: In ```RestExplainActionIT``` the variable names began with ```1``` and the corresponding index name ended with ```_1```. Let's keep it consistent and do:
```
indexName_1 = "{indexName}_1"
...
...
indexName_X = "{indexName}_X"
```

Except for cases like ```val indexName3 = "${testIndexName}_some_other_test_2"``` below which are exceptions.

jinsoor-amzn: this is set up so that it is easier to query index name with index pattern in following retry call.

`/$indexName1*` This way we and use the `indexName1` and can expect rest of the indices to be included.

qreshi: Oh sorry I didn't meant to include the ```_1``` on the variable name. I meant to start with ```1``` always on the variable names

PR #74, opened 2019-08-23
Clear ManagedIndexMetaData upon policy removal

qreshi:
*Issue #, if available:*

*Description of changes:*
Clears the ManagedIndexMetaData when a policy is removed from an index


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/UpdateManagedIndexMetaDataRequest.kt
```diff
@@ -24,38 +24,56 @@ import org.elasticsearch.common.io.stream.StreamOutput
...
+        indicesToAddManagedIndexMetaDataTo: List<Pair<Index, ManagedIndexMetaData>> = listOf(),
+        indicesToRemoveManagedIndexMetaDataFrom: List<Index> = listOf()
+    ) {
+        this.indicesToAddManagedIndexMetaDataTo = indicesToAddManagedIndexMetaDataTo
+        this.indicesToRemoveManagedIndexMetaDataFrom = indicesToRemoveManagedIndexMetaDataFrom
     }
 
     override fun validate(): ActionRequestValidationException? {
         var validationException: ActionRequestValidationException? = null
-        if (!this::listOfIndexMetadata.isInitialized) {
```
dbbaughe: Do we want && or || here?

qreshi: The constructor with parameters for ```UpdateManagedIndexMetaDataRequest``` will ensure that the lists are at least set to ```emptyList()``` which is how we invoke the request in our plugin. But if the values are being read in using ```readFrom()```, the lists could be null since the Transport class ```TransportUpdateManagedIndexMetaDataAction``` passes in the empty constructor as the ```Supplier```. Within that Transport class, I was referencing both lists (addTo/removeFrom) as if they were at most empty and not null so the validation was to confirm that assumption in the ```readFrom()``` case.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -343,6 +355,41 @@ class ManagedIndexCoordinator(
...
+    @OpenForTesting
+    @Suppress("TooGenericExceptionCaught")
+    suspend fun clearManagedIndexMetaData(indices: List<Index>) {
+        try {
+            val request = UpdateManagedIndexMetaDataRequest(indicesToRemoveManagedIndexMetaDataFrom = indices)
+
+            retryPolicy.retry(logger) {
+                val response: AcknowledgedResponse = client.suspendUntil { execute(UpdateManagedIndexMetaDataAction, request, it) }
+
+                if (!response.isAcknowledged) logger.error("Failed to remove ManagedIndexMetaData")
```
jinsoor-amzn: add index names to the error log.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/UpdateManagedIndexMetaDataRequest.kt
```diff
@@ -24,38 +24,56 @@ import org.elasticsearch.common.io.stream.StreamOutput
...
-            validationException = addValidationError("must specify index List for UpdateManagedIndexMetaData", validationException)
+        if (!this::indicesToAddManagedIndexMetaDataTo.isInitialized && !this::indicesToRemoveManagedIndexMetaDataFrom.isInitialized) {
```
jinsoor-amzn: I think this is not really necessary now. Previously there was no default list and the list always came from the constructor. Now that is not the case and we have default empty `listOf`. We should instead check if both of them are empty list then throw an exception .

qreshi: Added a check to ensure that at least one non-empty list is given but will also keep the check to validate that both lists are initialized for robustness.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/UpdateManagedIndexMetaDataRequest.kt
qreshi: After further discussion, we decided to use || to ensure both lists are initialized.

PR #93, opened 2019-08-30
Adds logic for enabling/disabling Index State Management 

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds support for enabling/disabling the management of indices

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexCoordinator.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexCoordinator.kt
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.indexstatemanagement
 
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementPlugin.Companion.INDEX_STATE_MANAGEMENT_INDEX
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.getClusterStateManagedIndexConfig
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.getManagedIndexMetaData
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.getPolicyID
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.retry
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.shouldCreateManagedIndexConfig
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -398,6 +398,15 @@ val ManagedIndexMetaData.isFailed: Boolean
         return false
     }
 
+val ManagedIndexMetaData.isPolicyCompleted: Boolean
+    get() {
+        // If policyCompleted is true, return true
+        if (this.policyCompleted == true) return true
```
jinsoor-amzn: this can be simplified to 

```
val ManagedIndexMetaData.isPolicyCompleted: Boolean
    get() = this.policyCompleted == true
```

also why do we need as separate variable? Can we achieve the same thing by just doing below?

```
if (policy.policyCompleted == true) {
    ...
}
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -172,13 +179,53 @@ class ManagedIndexCoordinator(
...
+        /*
+         * Iterate through all indices and create update requests to update the ManagedIndexConfig for indices that
+         * meet the following conditions:
+         *   1. Is being managed (has ManagedIndexMetaData)
+         *   2. Does not have a completed Policy
+         *   3. Does not have a failed Policy
+         */
+        val updateManagedIndicesRequests: List<DocWriteRequest<*>> = response.state.metaData.indices.mapNotNull {
+            val managedIndexMetaData = it.value.getManagedIndexMetaData()
+            if (!(managedIndexMetaData == null || managedIndexMetaData.isPolicyCompleted || managedIndexMetaData.isFailed)) {
```
jinsoor-amzn: Wouldn't this be true for even indices that are not managed at all? We need to check if the index is managed index.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
qreshi: True but I didn't like mixing predicate style ```isFailed``` with explicit boolean evaluation like ```policyCompleted == true```, it made the boolean statement less readable and inconsistent.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexCoordinator.kt
qreshi: This is checking if the ```ManagedIndexMetaData``` exists and that the job is neither completed nor failed. Since we now clear ```ManagedIndexMetaData``` when a policy is removed from an index, the ```ManagedIndexMetaData``` we fetch here should be for managed indices.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -398,6 +398,10 @@ val ManagedIndexMetaData.isFailed: Boolean
         return false
     }
 
+// Adding predicate extension to allow cleaner checks since policyCompleted is nullable
+val ManagedIndexMetaData.isPolicyCompleted: Boolean
```
dbbaughe: Can we just make policyCompleted non-nullable? We're parsing the value in ManagedIndexMetaData and then constructing the data class, we could just say something like `policyCompleted = policyCompleted ?: false` to make sure the value on the ManagedIndexMetaData is actually a boolean. And then where we check `if (policyCompleted != null) ...` we can just check if its true then show it.

@jinsoor-amzn Thoughts on that too?

qreshi: I agree with making ```policyCompleted``` non-nullable but given the scope of ```ManagedIndexMetaData```'s usage both in our logic and tests, let's do this change in a separate PR.

PR #96, opened 2019-09-24
Adds AllowList functionality

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds AllowList functionality
By default all actions are allowed
If PUT Policy API is called using an action not allowed then it will fail with 403 Forbidden
If AllowList is updated after a policy is created and a managed index using that policy tries to execute an action that is not allowed then it will fail w/ reason stating the action is not allowed

Chose AllowList over DenyList because forgetting to add an action to an AllowList is safer than forgetting to add an action to DenyList.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
@@ -34,6 +34,7 @@ import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedi
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedindexmetadata.StateMetaData
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings.Companion.ALLOW_LIST
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings.Companion.ALLOW_LIST_NONE
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings.Companion.DEFAULT_ISM_ENABLED
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings.Companion.DEFAULT_JOB_INTERVAL
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings.Companion.INDEX_STATE_MANAGEMENT_ENABLED
@@ -119,9 +120,8 @@ object ManagedIndexRunner : ScheduledJobRunner,
     private val updateMetaDataRetryPolicy = BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(250), 3)
     @Suppress("MagicNumber")
     private val errorNotificationRetryPolicy = BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(250), 3)
-    @Suppress("MagicNumber")
     private var jobInterval: Int = DEFAULT_JOB_INTERVAL
-    private var allowList: List<String> = emptyList()
+    private var allowList: List<String> = ALLOW_LIST_NONE
 
     fun registerClusterService(clusterService: ClusterService): ManagedIndexRunner {
         this.clusterService = clusterService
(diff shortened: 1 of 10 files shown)


dbbaughe:
BUILD SUCCESSFUL in 6m 15s

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -109,7 +109,7 @@ data class ManagedIndexMetaData(
             builder.endObject()
         }
 
-        if (policyRetryInfo != null && !transitionToExists) {
+        if (policyRetryInfo != null) {
```
jinsoor-amzn: Why are you removing this?

dbbaughe: When we set transitionTo, on the next execution we get the first action/firstStep of that next state. If that action is disallowed then we fail the policy (by just updating the metaData policy failed to true).

The metadata still has transitionTo set since we are not replacing it unless it actually transitioned. This means in the explain API the policy was failed but wasn't showing failed=true because a transitionTo existed.

This seems like it will happen for any policy failure on the first execution when transitioning, so we should probably always show PolicyRetryInfo regardless of transitionTo.

We still don't want to show action or state info because it refers to old state/action information (and if action/state information is updated it should mean the transitionTo was removed).

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/settings/ManagedIndexSettings.kt
```diff
@@ -109,5 +111,13 @@ class ManagedIndexSettings {
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         )
+
+        val ALLOW_LIST = Setting.listSetting(
+            "opendistro.index_state_management.allow_list",
+            ActionConfig.ActionType.values().toList().map { it.type },
+            Function.identity(),
```
qreshi: How is this ```Function.identity()``` used? Looking at the javadocs, it says that this argument is ```singleValueParser```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -119,6 +121,7 @@ object ManagedIndexRunner : ScheduledJobRunner,
     private val errorNotificationRetryPolicy = BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(250), 3)
     @Suppress("MagicNumber")
     private var jobInterval: Int = DEFAULT_JOB_INTERVAL
+    private var allowList: List<String> = emptyList()
```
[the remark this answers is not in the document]
dbbaughe: Defaulting to all actions in settings, but want to default to none in runner so that it goes from [none] -> [some] rather than [all] -> [some] just to be secure. Moved into variable though for both places.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -486,3 +486,24 @@ fun Policy.isSafeToChange(stateName: String?, newPolicy: Policy, changePolicy: C
 
     return true
 }
+
+/**
+ * Disallowed actions are ones that are not specified in the [ManagedIndexSettings.ALLOW_LIST] setting.
+ */
+fun Policy.getDisallowedActions(allowList: List<String>): List<String> {
+    val allowListSet = allowList.toSet()
+    val disallowedActions = mutableListOf<String>()
```
qreshi: NP: To be a little more space efficient, you can declare this as a ```mutableSetOf<String>``` instead so that duplicate actions are not added into ```disallowedActions``` and then convert it to a list at the end by doing ```disallowedActions.toList()```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -119,6 +121,7 @@ object ManagedIndexRunner : ScheduledJobRunner,
     private val errorNotificationRetryPolicy = BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(250), 3)
     @Suppress("MagicNumber")
```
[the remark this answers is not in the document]
dbbaughe: No, seemed leftover from previous changes. Removed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/settings/ManagedIndexSettings.kt
dbbaughe: Identity functions just return the value passed in. This is taken from examples in Elasticsearch using the Setting.listSetting. We aren't doing anything special with the strings in the list so just parsing and returning themselves should be fine.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/ManagedIndexUtils.kt
dbbaughe: Don't expect list to be large, will skip for now.

Issue #103, opened 2019-10-08
Implement allocation action


jgprogramming:
## Expected Behavior
Our opendistro cluster has two data node types: `hot` and `warm`. For each index we have varied retention times for hot and warm boxes so we using curator for this task gets tiresome. We'd like to use ILM for that task.

We expect that policy for allocate will look like:
```
{
    "allocation": {
        "require": { "box_type": "hot" }
    }
}
```

It's also a good idea to provide `include` and `exclude` filters.

## Current Behavior
allocate action is not implemented

dbbaughe:
Steps required:
[AttemptAllocationStep] -> Set the allocation index/shard filtering settings
[WaitForAllocationStep] -> Wait for shards to actually allocate to nodes

Settings:
"require"
"include"
"exclude"
"wait_for"

dabr-grapeup:
Hi @dbbaughe, I've a question about logic for WaitForAllocationStep, what API endpoints should be use to validate that? For me it's not clear how can I test if the current node matches allocation settings (require, include, exclude).

dabr-grapeup:
I've opened a PR with my changes regarding implementation of `AttemptAllocationStep`.

lolcatlolcat:
I'm assuming this is actively being worked on? Because we have similar requirements as well and would love to use ISM in our hot -> warm -> cold architecture in-lieu of the python script we have now.

jgprogramming:
Hi @alolita , any news regarding this feature? It's been over three months since this PR.

alexkleinfig:
It would be great.
I'am just waiting for this feature to migrate our classic 10To elk stack to opendistro

jgprogramming:
Hey @dbbaughe, can you take a look at this one? PR is stuck since October.

dbbaughe:
Hey @jgprogramming,

I will be picking this up as one of my next tasks. I believe it was only blocked because we need to add multi-node integration tests into the gradle tasks. Once I get that added then we'll just need the PR updated and we can merge.

Thanks

sgavrylenko:
Hi guys! 
As I understand, this feature is described on doc page (https://opendistro.github.io/for-elasticsearch-docs/docs/elasticsearch/cluster/) about cluster configuration, but not implemented yet?

From that page
`A popular approach is to configure your index templates to set the index.routing.allocation.require.temp value to hot. This way, Elasticsearch stores your most recent data on your hot nodes.

[...]

PR #106, opened 2019-10-30
Implemented allocation action which can be used in index life cycle management

dabr-grapeup:
*Issue #, if available:* #103

*Description of changes:* Implemented Allocation action which can be used in ILM policy, action sets allocation properties to given index settings.


dbbaughe:
Hi @dabr-grapeup,

Thanks for the PR!

I'm currently looking into an easy way to include multi-node integration tests in index management since we'll need it for this PR. Will look over the rest of the PR over the next couple days too.

arnitolog:
@dbbaughe have you had a chance to review this PR? I looking forward to this action.

dbbaughe:
Hey @dabr-grapeup and @arnitolog,

We have one main blocker for this PR which is the testing part. We need to setup our integration tests to run on a multi-node setup so we can accurately test the allocation part. If someone can pick that up it would be great, otherwise it will have to wait until we can allocate resources to that task.

Thanks,
Drew

dbbaughe:
Hi @dabr-grapeup,

If you have the time, this PR has been unblocked by
https://github.com/opendistro-for-elasticsearch/index-management/pull/254
https://github.com/opendistro-for-elasticsearch/index-management/pull/256

So we have a way to run multi node tests now. What we need to do is add an integration test for the Allocation action that successfully migrates the shards of the index to a different node and I believe you can use the node name or number (can't remember off top of my head.. I think it's one of those) to control the shard allocation. And we need the test to only conditionally run when it's being run in multi-node configuration which we can check with the [number of nodes property](https://github.com/opendistro-for-elasticsearch/index-management/pull/254/files#diff-2b3a8fdd53b04a91aad0b6cc3c0ee5c9R83).

If you get the chance to look this over, great! If not I'll try to pick it up shortly and write the test.

jgprogramming:
Hey @dbbaughe , I'm working on those tests, they should be ready tomorrow.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/allocation/AttemptAllocationStep.kt
```diff
@@ -0,0 +1,73 @@
...
+            val mutableInfo = mutableMapOf("message" to ERROR_MESSAGE)
+            val errorMessage = e.message
+            if (errorMessage != null) mutableInfo["cause"] = errorMessage
+            info = mutableInfo.toMap()
+        }
+    }
+
+    private fun buildSettings(): Settings {
+        val builder = Settings.builder()
+        config.require.forEach { (key, value) -> builder.put(SETTINGS_PREFIX + AllocationActionConfig.REQUIRE + "." + key, value) }
```
dbbaughe, line 45: Can the key values be absolutely anything? Or should we have some validation on them in the parser? i.e. can you have something like `"index.routing.allocation._?__": " "` or will it throw an exception?

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/AllocationActionIT.kt
```diff
@@ -0,0 +1,207 @@
...
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementRestTestCase
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.Policy
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.State
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.action.AllocationActionConfig
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.randomErrorNotification
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.waitFor
+import org.junit.Assume
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.*
```
dbbaughe, line 27: Avoid wildcard imports

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/AllocationActionIT.kt
```diff
@@ -0,0 +1,207 @@
...
+            val settings = getFlatSettings(indexName)
+            assertTrue(settings.containsKey("index.routing.allocation.require._name"))
+            assertEquals(actionConfig.require["_name"], settings["index.routing.allocation.require._name"])
+        }
+
+        // Third execution: Waits for allocation to complete, which will happen in this execution since index is small
+        updateManagedIndexConfigStartTime(managedIndexConfig)
+
+        waitFor {
+            assertEquals("integTest-1", getIndexShardNodes(indexName)[0])
```
dbbaughe, line 112: It's possible this started out on integTest-1 to begin with. I think we need to get the current value of the node its on and then migrate it to the other one and assert it or perhaps have two Allocation actions on the policy that swap it between both nodes and then add another updateManagedIndexConfigStartTime and assertion to show it was on both when required.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/AllocationActionIT.kt
```diff
@@ -0,0 +1,207 @@
...
+            val settings = getFlatSettings(indexName)
+            assertTrue(settings.containsKey("index.routing.allocation.exclude._name"))
+            assertEquals(actionConfig.exclude["_name"], settings["index.routing.allocation.exclude._name"])
+        }
+
+        // Third execution: Waits for allocation to complete, which will happen in this execution since index is small
+        updateManagedIndexConfigStartTime(managedIndexConfig)
+
+        waitFor {
+            assertNotEquals("integTest-0", getIndexShardNodes(indexName)[0])
```
dbbaughe, line 158: Same as above, it's possbiel this was not on integTest-0 to begin with

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/AllocationActionIT.kt
```diff
@@ -0,0 +1,207 @@
...
+            val settings = getFlatSettings(indexName)
+            assertTrue(settings.containsKey("index.routing.allocation.include._name"))
+            assertEquals(actionConfig.include["_name"], settings["index.routing.allocation.include._name"])
+        }
+
+        // Third execution: Waits for allocation to complete, which will happen in this execution since index is small
+        updateManagedIndexConfigStartTime(managedIndexConfig)
+
+        waitFor {
+            assertEquals("integTest-1", getIndexShardNodes(indexName)[0])
```
dbbaughe, line 204: Same as above

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/allocation/AttemptAllocationStep.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedindexmetadata.StepMetaData
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.step.Step
+import org.apache.logging.log4j.LogManager
+import org.elasticsearch.action.admin.indices.settings.put.UpdateSettingsRequest
+import org.elasticsearch.action.support.master.AcknowledgedResponse
+import org.elasticsearch.client.Client
+import org.elasticsearch.cluster.service.ClusterService
+import org.elasticsearch.common.settings.Settings
+
+class AttemptAllocationStep(
```
dbbaughe, line 15: Also saw the mappings had a wait_for which I'm assuming is a flag to optionally wait for the shards to finish migrating to their respective filtered nodes, but I don't see it used. Are we skipping it for the first implementation? That's fine if so, otherwise we would need another step (WaitForAllocation or something) that would periodically check the shard locations.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/allocation/AttemptAllocationStep.kt
jgprogramming: While it is not possible to configure key like `"index.routing.allocation._..__": " "` it is possible to configure `"index.routing.allocation._._._._.": " "` which is equally useless. As far as I know elastic doesn't provide one regex that we can match to so I've added a test that shows that any invalid key will result in a failed policy.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/AllocationActionIT.kt
jgprogramming: Done, I've added a check to see on which node index is allocated. I think it's a cleaner solution than allocating multiple times.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/allocation/AttemptAllocationStep.kt
jgprogramming: That's correct, it's a placeholder for future step.

jgprogramming:
@dbbaughe Thanks for your comments, let me know if there's anything else.

bowenlan-amzn:
Thanks!

PR #120, opened 2019-12-02
Create Build and Test Workflow for opendistro-1.2 branch

qreshi:
*Issue #, if available:*

*Description of changes:*
I'm suspecting that the workflow file needs to be present on each branch that will be running it since a recent PR I submitted did not trigger the workflow. Backporting the workflow commit from master to this branch to see if this is the case.


qreshi:
The workflow is failing because ```notifications``` has only been published to maven for 1.3. But it does look like the workflow triggered once the files were present in this branch. Will backport the workflow after the relevant```notifications``` versions has been published.

Issue #45, opened 2019-07-24
Implement Snapshot action


dbbaughe:

dbbaughe:
Steps:
[AttemptSnapshotStep] -> Call the snapshot API
- Don't fail this step if the snapshot API fails because of in progress
[WaitForSnapshotStep] -> Wait for snapshot to complete

Settings:
https://github.com/opendistro-for-elasticsearch/index-management/blob/master/src/main/resources/mappings/opendistro-ism-config.json#L181

dabr-grapeup:
@dbbaughe Hi, I'm working at the design of this feature and I have one doubt about steps, do we really need `WaitForSnapshotStep` as separate step instead of just option for `AttemptSnapshotStep`? As I see the Snapshot API supports `wait for completion` option, it would be much easier to implement this ways and it would be more natural in terms of Snapshot API.

dbbaughe:
Hi @dabr-grapeup,

The `WaitFor..` steps generally mimic `wait_for_completion` (or something similar if it doesn't exist natively). The reason we don't actually `wait_for` inside the `Attempt..` steps is because we don't want to have a single execution of a job actively waiting an extended period of time (which could be hours for some users indices). If there is some intermittent network issue and connection drops or node crashes then it's harder to recover from that in a single step that does both things, whereas if we break up the execution from the waiting then it becomes a bit more reliable and easier to retry from.

dabr-grapeup:
Hi @dbbaughe,
I've extended my changes in Snapshot implementation with WaitForSnapshotStep and created PR. However I'm having problems with building it on your CI server as well as reproducing the problem on any local machine I have as I'm able to run all tests and build artifact without any problems.
After running command to reproduce:
`./gradlew ':integTestRunner' --tests "com.amazon.opendistroforelasticsearch.indexstatemanagement.action.SnapshotActionIT.test basic" -Dtests.seed=5F6526A5066301FB -Dtests.security.manager=false -Dtests.locale=ms-BN -Dtests.timezone=America/Mexico_City -Dcompiler.java=12 -Druntime.java=12`
```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/Users/dabr/workspace/adidas/index-management/build.gradle' line: 45

* What went wrong:
A problem occurred evaluating root project 'opendistro-for-elasticsearch-index-management'.
> Failed to apply plugin [id 'elasticsearch.esplugin']
[...]

PR #135, opened 2020-01-15
Snapshot implementation

dabr-grapeup:
*Issue #, if available:* #45 

*Description of changes:*
Action for creating snapshot with two steps: attempt_snapshot and wait_for_snapshot


dbbaughe:
Pulled it down to check, seems like the `createRepository(repository)` in the snapshot IT test might be causing the issue or part of it.
Once I uncommented that line and ran the integration tests all the failures happened.

dbbaughe:
Hi @dabr-grapeup,

Just a quicks heads-up, will be able to take a look at this towards the end of this week.
Currently getting things ready for the ODFE 1.4 release.

Thanks,
Drew

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/SnapshotAction.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    private val stepNameToStep: LinkedHashMap<String, Step> = linkedMapOf(
+        AttemptSnapshotStep.name to attemptSnapshotStep,
+        WaitForSnapshotStep.name to waitForSnapshotStep
+    )
+
+    override fun getSteps(): List<Step> = stepNameToStep.values.toList()
+
+    @Suppress("ReturnCount")
+    override fun getStepToExecute(): Step {
```
dbbaughe: Seems like this doesn't handle the case where the step is still on waitForSnapshotStep? Looks like it might return attemptSnapshotStep instead.
i.e. the snapshot is taking a while to complete for a large index and it will take multiple executions for waitFor to successfully complete.

Also if the step was on AttempSnapshotStep but failed/was retried. Seems like it will move forward to waitForSnapshotStep?
I'm assuming we only want to move to waitFor step current step is AttemptSnapshotStep and stepMetaData.stepStatus has a status of COMPLETED.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
...
+    private var info: Map<String, Any>? = null
+
+    @Suppress("TooGenericExceptionCaught")
+    override suspend fun execute() {
+        try {
+            logger.info("Executing snapshot on ${managedIndexMetaData.index}")
+            val createSnapshotRequest = CreateSnapshotRequest()
+                .indices(managedIndexMetaData.index)
+                .snapshot(config.snapshot)
+                .repository(config.repository)
```
dbbaughe: Can we explicitly set the waitForCompletion to false just in case they change the default behavior in the future.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
...
+            if (config.includeGlobalState != null) {
+                createSnapshotRequest.includeGlobalState(config.includeGlobalState)
+            }
+
+            val response: CreateSnapshotResponse = client.admin().cluster().suspendUntil { createSnapshot(createSnapshotRequest, it) }
```
dbbaughe: What's response type for a different snapshot already in progress?
If it's FAILED we probably do not want to treat it as failed since that's an OK state to be in. We should treat such failures as noops and just try again on the next execution. We have a CONDITION_NOT_MET step status we could use for such cases... not sure if it's the right meaning for this case though. Maybe a new NOOP step status?

Any other "failures" that really shouldn't be failures? We have a best effort skip when cluster status is RED in the runner logic. But, technically that check could be green and in-between that check and this step the cluster goes red. Do individual snapshots of single indices go through if cluster is red?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
...
+                RestStatus.OK -> {
+                    stepStatus = StepStatus.COMPLETED
+                    info = mapOf("message" to "Snapshot created for index: ${managedIndexMetaData.index}")
+                }
+                else -> {
+                    stepStatus = StepStatus.FAILED
+                    info = mapOf("message" to "There was an error during snapshot creation for index: ${managedIndexMetaData.index}")
+                }
+            }
+        } catch (e: Exception) {
```
dbbaughe: Do failures such as no repository exists, duplicate snapshot name, etc. throw an exception and get caught here or do they return successfully with RestStatus.INTERNAL_SERVER_ERROR? If it's the latter, we probably still want to capture the relevant error message and add it to the info object above.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    private val logger = LogManager.getLogger(javaClass)
+    private var stepStatus = StepStatus.STARTING
+    private var info: Map<String, Any>? = null
+
+    @Suppress("TooGenericExceptionCaught")
+    override suspend fun execute() {
+        try {
+            logger.info("Executing snapshot on ${managedIndexMetaData.index}")
+            val createSnapshotRequest = CreateSnapshotRequest()
```
dbbaughe: I think this can take some custom metadata stored with the snapshot. Could be useful to add some metadata so user knows this snapshot was generated by the Index Management plugin for this index. Maybe even the policy_id used, current state, etc.? What do you think?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
...
+                .indices(managedIndexMetaData.index)
+                .snapshot(config.snapshot)
```
dbbaughe: Should we add a unique hash to the end of the snapshot name? Since duplicate snapshot names are not allowed this depends on how the user defined their policy and snapshot action.

If it's a linear flow, i.e. hot -> warm -> cold -> delete where delete has 1 snapshot executed then it seems fine.

But, if it's a circular flow where it continuously takes snapshots then they have to use date math expressions in the name. If it executes fast enough though, they could be taking multiple snapshots per day and they need to make sure to use the correct date math expression. Might be easier for the user if we just take the user defined snapshot name and append a unique hash at the end? So they can still reference `<there_concrete_name>*`.

Thoughts?

JavierAbrego:
Hi @dbbaugh, thanks for your comments, we have fixed some of the issues that you commented, let me know if you spot something else.

Thanks

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/SnapshotAction.kt
```diff
@@ -0,0 +1,36 @@
+package com.amazon.opendistroforelasticsearch.indexstatemanagement.action
```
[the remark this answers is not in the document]
JavierAbrego: Done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
+package com.amazon.opendistroforelasticsearch.indexstatemanagement.step.snapshot
```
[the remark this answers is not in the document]
JavierAbrego: Done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/WaitForSnapshotStep.kt
```diff
@@ -0,0 +1,61 @@
+package com.amazon.opendistroforelasticsearch.indexstatemanagement.step.snapshot
```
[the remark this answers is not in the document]
JavierAbrego: Done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
JavierAbrego: Done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
JavierAbrego: we updated the code to handle these failures as well and change the status to CONDITION_NOT_MET while it's in progress

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
JavierAbrego: Done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
JavierAbrego: Added

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
JavierAbrego: We've added the DateTime to the snapshot name

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,75 @@
...
+    @Suppress("TooGenericExceptionCaught")
+    override suspend fun execute() {
+        try {
+            logger.info("Executing snapshot on ${managedIndexMetaData.index}")
+            val createSnapshotRequest = CreateSnapshotRequest()
+                .indices(managedIndexMetaData.index)
+                .snapshot(config.snapshot)
+                .repository(config.repository)
+            if (config.includeGlobalState != null) {
+                createSnapshotRequest.includeGlobalState(config.includeGlobalState)
```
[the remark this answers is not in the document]
JavierAbrego: I've changed the logic

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/SnapshotAction.kt
```diff
@@ -0,0 +1,36 @@
...
+class SnapshotAction(
+    clusterService: ClusterService,
+    client: Client,
+    managedIndexMetaData: ManagedIndexMetaData,
+    config: SnapshotActionConfig
+) : Action(ActionType.SNAPSHOT, config, managedIndexMetaData) {
+    private val attemptSnapshotStep = AttemptSnapshotStep(clusterService, client, config, managedIndexMetaData)
+    private val waitForSnapshotStep = WaitForSnapshotStep(clusterService, client, config, managedIndexMetaData)
+
+    private val stepNameToStep: LinkedHashMap<String, Step> = linkedMapOf(
```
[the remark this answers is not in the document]
JavierAbrego: removed

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/SnapshotAction.kt
JavierAbrego: Updated, I've changed the logic to suit these cases.

dbbaughe:
Thanks @dabr-grapeup and @JavierAbrego,

Will pull down and take another look at it. The main thing I see from GitHub is the global state parameter which I think we'll need to remove. But, let me confirm it has the negative effect we're trying to avoid.

@jinsoor-amzn @qreshi Can one of you also take a look as second reviewer

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,115 @@
...
+import org.elasticsearch.action.admin.cluster.snapshots.create.CreateSnapshotRequest
+import org.elasticsearch.action.admin.cluster.snapshots.create.CreateSnapshotResponse
+import org.elasticsearch.client.Client
+import org.elasticsearch.cluster.service.ClusterService
+import org.elasticsearch.rest.RestStatus
+import org.elasticsearch.snapshots.ConcurrentSnapshotExecutionException
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.util.*
```
jinsoor-amzn, line 33: avoid using `*` import.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementRestTestCase.kt
```diff
@@ -18,36 +18,29 @@ package com.amazon.opendistroforelasticsearch.indexstatemanagement
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementPlugin.Companion.INDEX_STATE_MANAGEMENT_HISTORY_TYPE
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementPlugin.Companion.INDEX_STATE_MANAGEMENT_INDEX
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementPlugin.Companion.POLICY_BASE_URI
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.ChangePolicy
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.ManagedIndexConfig
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.ManagedIndexMetaData
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.Policy
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.*
```
jinsoor-amzn, line 21: avoid using `*` import.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementRestTestCase.kt
```diff
@@ -18,36 +18,29 @@ package com.amazon.opendistroforelasticsearch.indexstatemanagement
...
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedindexmetadata.StateMetaData
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.resthandler.RestExplainAction
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.settings.ManagedIndexSettings
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.FAILED_INDICES
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.FAILURES
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.UPDATED_INDICES
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.util._ID
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.util._PRIMARY_TERM
-import com.amazon.opendistroforelasticsearch.indexstatemanagement.util._SEQ_NO
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.*
```
jinsoor-amzn, line 28: avoid using `*` import.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/action/SnapshotActionIT.kt
```diff
@@ -0,0 +1,70 @@
...
+
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementRestTestCase
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.Policy
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.State
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.action.SnapshotActionConfig
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.randomErrorNotification
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.waitFor
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.*
```
jinsoor-amzn, line 26: avoid using `*` import.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -0,0 +1,115 @@
...
+                    mutableInfo["message"] = "Snapshot created for index: ${managedIndexMetaData.index}"
+                }
+                else -> {
+                    stepStatus = StepStatus.FAILED
+                    mutableInfo["message"] = "There was an error during snapshot creation for index: ${managedIndexMetaData.index}"
+                    mutableInfo["cause"] = response.toString()
+                }
+            }
+            info = mutableInfo.toMap()
+        } catch (e: ConcurrentSnapshotExecutionException) {
```
jinsoor-amzn, line 88: Since this step is considered `idempotent` how do we know if the existing snapshot creation in progress is from this step or from user triggered snapshot?

We should consider the StepStatus to be `completed` if the snapshot is already in progress by this step and move on to `WaitForSnapshot`

dbbaughe: @jinsoor-amzn ConcurrentSnapshotExecutionException is thrown when any snapshot is in progress, possibly one that does not include this index.

Did not see the isIdempotent was set to `true` here, good catch.

We probably want to change it to `false`. Considering this instead causes a `new` snapshot to be created even if the previous one was already kicked off and completed I'm not sure I'd consider it idempotent.. although since these are technically just diffs built on top of each other I'm not sure it would really hurt.

For clarification  @dabr-grapeup and @JavierAbrego, we set steps to idempotent when it really doesn't matter if it gets called again if the previous execution failed to fully complete. Things like setting the index settings (readonly, readwrite) can be called multiple times with the same response. Same with close/open, these can be called multiple times with no side effects. Calling close on an already closed index simply returns a 200 ack:true. Calling snapshot multiple times can end up with multiple snapshots being created which might be OK for some people and not OK for others.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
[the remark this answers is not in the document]
dbbaughe: I did some testing with `include_global_state` and it's not something we can include at this point in time.

It's including our custom index metadata in the snapshot so when you restore it actually resets the index being managed back to that point in time when the snapshot occurred.

@dabr-grapeup @JavierAbrego Could you remove `include_global_state` from the steps/action/parsing so it's not included for now. This can be a followup on how we can include this in the snapshot action. Either we need to find a way to make sure the ISM custom index metadata is not included in the global state with the snapshot or we need to consider moving the metadata out of the cluster state and possibly into the config index.

alolita:
Hi @dabr-grapeup @JavierAbrego - There are some changes our reviewers have requested. Could you please take a look and make the requested changes so that we can merge. Please let us know if you have any questions! Thanks for your patience on this PR :-)

JavierAbrego:
Hello @alolita @jinsoor-amzn  @dbbaughe, thanks for your comments, I've pushed the changes. 
Let me know if you spot something else : )

dbbaughe:
Hey @JavierAbrego,

Thanks for the quick update, I believe there was just one comment missed [here](https://github.com/opendistro-for-elasticsearch/index-management/pull/135#discussion_r413272173). I tested the include_global_state option and it's not something we can include in this first release as it doesn't work well with ISM's metadata in the cluster state. We'll need to address it in a follow up release. Should be able to just remove the logic where it's added to the create snapshot request. Once we get that removed we should be able to merge.

JavierAbrego:
Sure @dbbaughe I just removed the logic. :+1:

PR #165, opened 2020-03-30
Adds isIdempotent method to each step and updates ManagedIndexRunner …

dbbaughe:
…to use it

*Issue #, if available:*

*Description of changes:*
Before every execution of a step, we first update the step_status in cluster state to [StepStatus.STARTING] to signal that work is about to be done for the managed index. The step then attempts to do work by calling execute, and finally updates the step_status with the results of that work ([StepStatus]).

If we ever start an execution with a step_status of [StepStatus.STARTING] it means we failed to update the step_status after calling the execute function. Since we do not know if the execution was a noop, failed, or completed then we can't always assume it's safe to just retry it (e.g. calling force merge multiple times in a row). This means hat final update is a failure point that can't be retried and when multiplied by # of executions it leads to a lot of chances over time for random network failures, timeouts, etc.

To get around this every step should have an [isIdempotent] method to signal if it's safe to retry this step for such failures.

Choices for isIdempotent value for each step:
- CloseAction
  - AttemptCloseStep: **YES**
- DeleteAction
  - AttemptDeletestep: **YES**
- ForceMergeAction
  - AttemptForceMergeStep: **NO**
  - AttemptSetReadOnlyStep: **YES**
  - WaitForForceMergeStep: **YES**
- NotificationAction
  - AttemptNotificationStep: **NO**
- OpenAction
  - AttemptOpenStep: **YES**
- ReadOnlyAction
  - SetReadOnlyStep: **YES**
- ReadWriteAction
  - SetReadWriteStep: **YES**
- ReplicaCountAction
  - AttemptSetReplicaCountStep: **YES**
- RolloverAction
  - AttemptRolloverStep: **NO**
- TransitionsAction
  - AttemptTransitionStep: **YES**

Unfortunately cannot think of an easy deterministic way to test this directly from our REST integration tests as this involves modifying the cluster state into a bad state or something forcing it to fail that last update of cluster state. Most likely will need to utilize the ESIntegTestCase for this, but that will be done as a follow up PR so we can get this out faster.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
@@ -265,10 +265,14 @@ object ManagedIndexRunner : ScheduledJobRunner,
         }
 
         if (managedIndexMetaData.stepMetaData?.stepStatus == Step.StepStatus.STARTING) {
-            val info = mapOf("message" to "Previous action was not able to update IndexMetaData.")
-            val updated = updateManagedIndexMetaData(managedIndexMetaData.copy(policyRetryInfo = PolicyRetryInfoMetaData(true, 0), info = info))
-            if (updated) disableManagedIndexConfig(managedIndexConfig)
-            return
+            val isIdempotent = step?.isIdempotent()
+            logger.info("Previous execution failed to update step status, isIdempotent=$isIdempotent")
+            if (isIdempotent != true) {
+                val info = mapOf("message" to "Previous action was not able to update IndexMetaData.")
+                val updated = updateManagedIndexMetaData(managedIndexMetaData.copy(policyRetryInfo = PolicyRetryInfoMetaData(true, 0), info = info))
+                if (updated) disableManagedIndexConfig(managedIndexConfig)
+                return
+            }
         }
 
         // If this action is not allowed and the step to be executed is the first step in the action then we will fail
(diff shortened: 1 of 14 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -265,10 +265,14 @@ object ManagedIndexRunner : ScheduledJobRunner,
...
         if (managedIndexMetaData.stepMetaData?.stepStatus == Step.StepStatus.STARTING) {
-            val info = mapOf("message" to "Previous action was not able to update IndexMetaData.")
-            val updated = updateManagedIndexMetaData(managedIndexMetaData.copy(policyRetryInfo = PolicyRetryInfoMetaData(true, 0), info = info))
-            if (updated) disableManagedIndexConfig(managedIndexConfig)
-            return
+            val isIdempotent = step?.isIdempotent()
+            logger.info("Previous execution failed to update step status, isIdempotent=$isIdempotent")
+            if (isIdempotent != true) {
+                val info = mapOf("message" to "Previous action was not able to update IndexMetaData.")
+                val updated = updateManagedIndexMetaData(managedIndexMetaData.copy(policyRetryInfo = PolicyRetryInfoMetaData(true, 0), info = info))
```
zengyan-amazon: what if this update index meta data fails?

dbbaughe: We would only reach this update if the step is not idempotent.
And if this update fails, then no matter what it will return two lines below.
So it just means on the next execution it would go through the runner logic again, see the step still in STARTING, see the step is not idempotent, attempt to update the metadata to failed again, disable the job if the update was successful, and return.

PR #172, opened 2020-04-03
Delete and close failing during snapshot in progress

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Currently DeleteAction and CloseAction will treat a SnapshotInProgressException as a failure that will consume action retries or fail if none.

These should be treated as noops instead and allowed to retry automatically.

Checked [here](https://github.com/elastic/elasticsearch/search?q=SnapshotInProgressException+path%3A%2Fserver%2Fsrc%2Fmain&unscoped_q=SnapshotInProgressException+path%3A%2Fserver%2Fsrc%2Fmain) to see where this exception can come from and only found Delete and Close.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
@@ -25,6 +25,7 @@ import org.elasticsearch.action.admin.indices.close.CloseIndexRequest
 import org.elasticsearch.action.admin.indices.close.CloseIndexResponse
 import org.elasticsearch.client.Client
 import org.elasticsearch.cluster.service.ClusterService
+import org.elasticsearch.snapshots.SnapshotInProgressException
 
 class AttemptCloseStep(
     val clusterService: ClusterService,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
```diff
@@ -41,21 +42,27 @@ class AttemptCloseStep(
...
-                info = mapOf("message" to "Failed to close index: ${managedIndexMetaData.index}")
+                info = mapOf("message" to "Failed to close index")
             }
+        } catch (e: SnapshotInProgressException) {
+            logger.warn("Failed to close index [index=$index] with snapshot in progress")
+            stepStatus = StepStatus.CONDITION_NOT_MET
+            info = mapOf("message" to "Index had snapshot in progress, retrying closing")
         } catch (e: Exception) {
-            logger.error("Failed to set index to close [index=${managedIndexMetaData.index}]", e)
+            logger.error("Failed to set index to close [index=$index]", e)
```
jinsoor-amzn, line 65: we can remove `[index=$index]` here also.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/AttemptCloseStepTests.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.elasticsearch.test.ESTestCase
+import kotlin.IllegalArgumentException
+
+class AttemptCloseStepTests : ESTestCase() {
+
+    private val clusterService: ClusterService = mock()
+
+    fun `test close step sets step status to completed when successful`() {
+        val closeIndexResponse = CloseIndexResponse(true, true, listOf())
+        val client = getClient(getAdminClient(getIndicesAdminClient(closeIndexResponse, null)))
```
jinsoor-amzn, line 43: Other tests we did not mock a "good" path and try to use as much native ES response as possible. Do we want to follow in this test? Or if do we want to start change other tests to do mocks?

I think we should be consistent through out other tests and only mock the ones that are difficult to reproduce.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/close/AttemptCloseStep.kt
dbbaughe: Hmm why? We won't know what index it was for then.

jinsoor-amzn: I thought this was a message for the metadata. You are right we do want to write it in error log.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/AttemptCloseStepTests.kt
dbbaughe: Discussed offline, this is just unit tests w/ mocks to tests the step status/outcome based off response. It's fine to have this test and another bigger integration test that does more.

PR #180, opened 2020-04-06
Update test-and-build-workflow.yml

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Some of our PRs get stuck with required status checks, most likely because the name of the status check is used as the key and older branches will expect a JDK 13 check to pass or newer branches were expecting a JDK 12 to pass. This removes the variable from the name.

### DIFF
--- a/.github/workflows/test-and-build-workflow.yml
+++ b/.github/workflows/test-and-build-workflow.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         java: [13]
     # Job name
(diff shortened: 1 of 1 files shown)


dbbaughe:
Yep, seems like the (#) is gone now from the name:
https://github.com/opendistro-for-elasticsearch/index-management/pull/180/checks?check_run_id=565625532

Screen Shot 2020-04-06 at 1 57 48 PM

PR #198, opened 2020-04-10
schema mapping change

jinsoor-amzn:
*Issue #, if available:*
N/A

*Description of changes:*
Update schema version for IndexManagementConfig mapping if available.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementIndices.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementIndices.kt
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.indexstatemanagement
 
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.IndexStateManagementPlugin.Companion.INDEX_STATE_MANAGEMENT_INDEX
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.elasticapi.suspendUntil
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.IndexUtils
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.OpenForTesting
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util._DOC
 import org.apache.logging.log4j.LogManager
@@ -27,6 +28,7 @@ import org.elasticsearch.action.admin.indices.create.CreateIndexRequest
 import org.elasticsearch.action.admin.indices.create.CreateIndexResponse
 import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsRequest
 import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsResponse
+import org.elasticsearch.action.support.master.AcknowledgedResponse
 import org.elasticsearch.client.Client
 import org.elasticsearch.client.IndicesAdminClient
 import org.elasticsearch.cluster.service.ClusterService
@@ -40,15 +42,21 @@ class IndexStateManagementIndices(
 
     private val logger = LogManager.getLogger(javaClass)
 
-    val indexStateManagementMappings = javaClass.classLoader.getResource("mappings/opendistro-ism-config.json").readText()
-
-    val indexStateManagementHistoryMappings = javaClass.classLoader.getResource("mappings/opendistro-ism-history.json").readText()
-
-    fun initIndexStateManagementIndex(actionListener: ActionListener<CreateIndexResponse>) {
+    fun checkAndUpdateISMConfigIndex(actionListener: ActionListener<AcknowledgedResponse>) {
         if (!indexStateManagementIndexExists()) {
             val indexRequest = CreateIndexRequest(INDEX_STATE_MANAGEMENT_INDEX)
                     .mapping(_DOC, indexStateManagementMappings, XContentType.JSON)
-            client.create(indexRequest, actionListener)
+            client.create(indexRequest, object : ActionListener<CreateIndexResponse> {
+                override fun onFailure(e: Exception) {
+                    actionListener.onFailure(e)
+                }
+
+                override fun onResponse(response: CreateIndexResponse) {
+                    actionListener.onResponse(response)
+                }
+            })
+        } else {
+            IndexUtils.checkAndUpdateConfigIndexMapping(clusterService.state(), client, actionListener)
         }
     }
 
@@ -62,7 +70,7 @@ class IndexStateManagementIndices(
         if (indexStateManagementIndexExists()) return true
 
         return try {
-            val response: CreateIndexResponse = client.suspendUntil { initIndexStateManagementIndex(it) }
+            val response: AcknowledgedResponse = client.suspendUntil { checkAndUpdateISMConfigIndex(it) }
             if (response.isAcknowledged) {
                 return true
             }
@@ -118,5 +126,8 @@ class IndexStateManagementIndices(
         const val HISTORY_WRITE_INDEX_ALIAS = "$HISTORY_INDEX_BASE-write"
         const val HISTORY_INDEX_PATTERN = "<$HISTORY_INDEX_BASE-{now/d{yyyy.MM.dd}}-1>"
         const val HISTORY_ALL = "$HISTORY_INDEX_BASE*"
+
+        val indexStateManagementMappings = javaClass.classLoader.getResource("mappings/opendistro-ism-config.json").readText()
+        val indexStateManagementHistoryMappings = javaClass.classLoader.getResource("mappings/opendistro-ism-history.json").readText()
     }
 }
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestChangePolicyAction.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestChangePolicyAction.kt
@@ -26,6 +26,7 @@ import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.coordina
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.FAILED_INDICES
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.FAILURES
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.FailedIndex
+import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.IndexUtils
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.UPDATED_INDICES
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.isSafeToChange
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.updateManagedIndexRequest
@@ -40,6 +41,7 @@ import org.elasticsearch.action.get.GetResponse
 import org.elasticsearch.action.search.SearchRequest
 import org.elasticsearch.action.search.SearchResponse
 import org.elasticsearch.action.support.IndicesOptions
+import org.elasticsearch.action.support.master.AcknowledgedResponse
 import org.elasticsearch.client.node.NodeClient
 import org.elasticsearch.cluster.service.ClusterService
 import org.elasticsearch.common.Strings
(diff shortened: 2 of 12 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementIndices.kt
```diff
@@ -40,15 +41,15 @@ class IndexStateManagementIndices(
...
-    val indexStateManagementHistoryMappings = javaClass.classLoader.getResource("mappings/opendistro-ism-history.json").readText()
-
     fun initIndexStateManagementIndex(actionListener: ActionListener<CreateIndexResponse>) {
         if (!indexStateManagementIndexExists()) {
             val indexRequest = CreateIndexRequest(INDEX_STATE_MANAGEMENT_INDEX)
-                    .mapping(_DOC, indexStateManagementMappings, XContentType.JSON)
+                    .mapping(_DOC, indexStateManagementMappings(), XContentType.JSON)
             client.create(indexRequest, actionListener)
+        } else {
+            IndexUtils.updateIndexMapping(INDEX_STATE_MANAGEMENT_INDEX,
```
dbbaughe, line 50: Curious what behavior is when this updateIndexMapping is called twice in a row from different nodes/requests?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
```diff
@@ -0,0 +1,111 @@
...
+
+        @JvmStatic
+        fun updateIndexMapping(
+            index: String,
+            mapping: String,
+            clusterState: ClusterState,
+            client: IndicesAdminClient,
+            actionListener: ActionListener<out AcknowledgedResponse>
+        ) {
+            actionListener as ActionListener<AcknowledgedResponse>
```
dbbaughe, line 99: Was there an issue with setting this in the parameter list?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.elasticsearch.common.xcontent.NamedXContentRegistry
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentType
+
+class IndexUtils {
+    companion object {
+        @Suppress("ObjectPropertyNaming")
+        const val _META = "_meta"
+        const val SCHEMA_VERSION = "schema_version"
+        const val NO_SCHEMA_VERSION = 1L
```
dbbaughe, line 35: nit: Should we change this to DEFAULT_SCHEMA_VESRION? That seems to describe the value better, i.e. if there is no schema version specified we use the default schema version.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementIndices.kt
```diff
@@ -118,5 +119,8 @@ class IndexStateManagementIndices(
         const val HISTORY_WRITE_INDEX_ALIAS = "$HISTORY_INDEX_BASE-write"
         const val HISTORY_INDEX_PATTERN = "<$HISTORY_INDEX_BASE-{now/d{yyyy.MM.dd}}-1>"
         const val HISTORY_ALL = "$HISTORY_INDEX_BASE*"
+
+        fun indexStateManagementMappings() = javaClass.classLoader.getResource("mappings/opendistro-ism-config.json").readText()
```
dbbaughe, line 123: Is this loading the mappings now every time this function is called?
Seems like before it was done just once on the class initialization and now it'll be loading from disk every time we need to compare mappings? Or does getResources cache/memoize anything internally?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
```diff
@@ -0,0 +1,111 @@
...
+
+        var indexManagementSchemaVersion: Long
+            private set
+        var indexManagementHistoryVersion: Long
+            private set
+
+        init {
```
dbbaughe, line 42: A bit confused by this classes logic. It seems like when it's initialized it's already parsing the mappings from disk and setting the schema versions, but then also during initIndexStateManagementIndex we pass in the mappings again from disk to updateIndexMapping  and call the getSchemaVersion again by parsing the mappings to do the comparison?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
```diff
@@ -0,0 +1,111 @@
...
+                val metaData = indexMapping[_META] as HashMap<*, *>
+                if (metaData.containsKey(SCHEMA_VERSION)) {
+                    oldVersion = (metaData[SCHEMA_VERSION] as Int).toLong()
+                }
+            }
+            return newVersion > oldVersion
+        }
+
+        @JvmStatic
+        fun updateIndexMapping(
```
dbbaughe, line 92: nit: Should this method just do what its name implies? Just update the mappings, and leave the shouldUpdateIndex check to be called by the caller of this method?

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -1,4 +1,7 @@
 {
+  "_meta" : {
+    "schema_version": 2
```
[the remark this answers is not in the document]
dbbaughe: I think this can be ignored, seems like we used a default of 1 when no schema version was set.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestIndexPolicyAction.kt
```diff
@@ -140,6 +141,8 @@ class RestIndexPolicyAction(
         }
 
         private fun putPolicy() {
+            newPolicy.copy(schemaVersion = IndexUtils.indexManagementSchemaVersion)
```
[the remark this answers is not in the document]
jinsoor-amzn: the mapping logic does not need to be that complicated. All of the logic is happening in `initIndexStateManagementIndex`.

1. **does not exist -> create index w/ mappings**  This is current behavior.
2. **exists w/ lower schema number -> update index w/ new mappings**  This is new implemented behavior.
3. **exists w/ higher schema number -> error? when would this happen, only if downgrading plugin version?** In this case we do not do anything. No error and no update. We just try to use existing higher schema version. This can happen if some node has higher version of ISM plugin but should be highly unlikely.
4. **exists w/ same schema number -> proceed with normal logic** This is current behavior.
5. **exists w/ no schema number -> error? somehow dynamic mappings were created** No current json does not have schema number so it is considered `NO_SCHEMA_VERSION` and compare with new schema version.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestChangePolicyAction.kt
```diff
@@ -113,7 +117,20 @@ class RestChangePolicyAction(controller: RestController, val clusterService: Clu
             if (!response.isExists || response.isSourceEmpty) {
                 return channel.sendResponse(BytesRestResponse(RestStatus.NOT_FOUND, "Could not find policy=${changePolicy.policyID}"))
             }
+            this.response = response
+            IndexUtils.updateIndexMapping(INDEX_STATE_MANAGEMENT_INDEX,
```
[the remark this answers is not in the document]
jinsoor-amzn: because we do not need to update the mapping if policy does not exist. Only if the user update ISM plugin version and their first call to ISM is to change policy then we want to ensure updating index mapping.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementIndices.kt
jinsoor-amzn: Then the mapping will be updated twice. Unless user is using different ISM plugin version it would just be updating the mapping to same version twice.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
jinsoor-amzn: Do you mean use `ActionListener<AcknowledgedResponse>` as param itself?

The reason I did this was because this function gets called by other type of actionListener `ActionListener<CreateIndexResponse>` in `initIndexStateManagementIndex`. So I had to use generic mapping to map `CreateIndexResponse` into `AcknoledgeResponse` to be used in `PutMappingRequest`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
jinsoor-amzn: That will lead to duplicate code outside for other places to check `shouldUpdateIndex`. I can rename this function to something like `updateIndexMappingIfNeeded` ? But I think `updateIndexMapping` is fine.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
jinsoor-amzn: Yes these mapping normally will be identical to what has already parsed to set the schema version but this function to pass-in mapping helps for integration test. It helps to change mapping schema version.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementIndices.kt
jinsoor-amzn: Changed to `val`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
jinsoor-amzn: changed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/util/IndexUtils.kt
```diff
@@ -0,0 +1,123 @@
...
+        @OpenForTesting
+        fun checkAndUpdateIndexMapping(
+            index: String,
+            schemaVersion: Long,
+            mapping: String,
+            clusterState: ClusterState,
+            client: IndicesAdminClient,
+            actionListener: ActionListener<AcknowledgedResponse>
+        ) {
+            if (clusterState.metaData.indices.containsKey(index)) {
```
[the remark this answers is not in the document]
jinsoor-amzn: done.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/IndexStateManagementRestTestCase.kt
```diff
@@ -516,6 +517,40 @@ abstract class IndexStateManagementRestTestCase : ESRestTestCase() {
...
     }
 
+    protected fun assertIndexExists(index: String) {
+        val response = client().makeRequest("HEAD", index)
+        assertEquals("Index $index does not exist.", RestStatus.OK, response.restStatus())
+    }
+
+    protected fun assertIndexDoesNotExist(index: String) {
+        val response = client().makeRequest("HEAD", index)
+        assertEquals("Index $index does not exist.", RestStatus.NOT_FOUND, response.restStatus())
```
qreshi, line 527: NP: Assertion message here should be printing out "Index ```does``` exist" if the response status is not ```RestStatus.NOT_FOUND``` I believe.

jinsoor-amzn: fixed.

PR #209, opened 2020-04-29
Switches UpdateManagedIndexMetaData to batch tasks using custom executor

dbbaughe:
*Issue #, if available:*

*Description of changes:*
The task updating the cluster state for ISM were being executed and applied individually. This moves us to our own executor that will batch the tasks (of ISM type) together so they can be processed faster.

From testing this has made significant improvements in the timeouts experienced on clusters with large # of managed indices.

Tested on two domains with 10 data nodes (m4.xlarge), 3 dedicated masters (m4.large) with 5,000 Managed Indices each.

The current ISM would not be able to initialize all managed indices and would take hours to get through the queue while eventually timing out 3.5k of them until it was able to manage 1.5k without timeouts.

ISM with these changes was able to process all 5k managed indices right away and was able bring down pending_tasks queue to comparably low # with no timeouts. Was able to stress test further by reducing job_interval to 1 minute (5k jobs running per minute, ~10k cluster state updates per minute) and this has been running with no failures/timeouts overnight. Not recommending to reduce to 1 minute though until other improvements, but it's good to see impact.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/TransportUpdateManagedIndexMetaDataAction.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/TransportUpdateManagedIndexMetaDataAction.kt
@@ -28,19 +28,25 @@ import org.elasticsearch.action.support.ActionFilters
 import org.elasticsearch.action.support.master.AcknowledgedResponse
 import org.elasticsearch.action.support.master.TransportMasterNodeAction
 import org.elasticsearch.client.Client
-import org.elasticsearch.cluster.AckedClusterStateUpdateTask
 import org.elasticsearch.cluster.ClusterState
+import org.elasticsearch.cluster.ClusterStateTaskConfig
+import org.elasticsearch.cluster.ClusterStateTaskExecutor
+import org.elasticsearch.cluster.ClusterStateTaskExecutor.ClusterTasksResult
+import org.elasticsearch.cluster.ClusterStateTaskListener
 import org.elasticsearch.cluster.block.ClusterBlockException
 import org.elasticsearch.cluster.block.ClusterBlockLevel
 import org.elasticsearch.cluster.metadata.IndexMetaData
 import org.elasticsearch.cluster.metadata.IndexNameExpressionResolver
 import org.elasticsearch.cluster.metadata.MetaData
 import org.elasticsearch.cluster.service.ClusterService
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/TransportUpdateManagedIndexMetaDataAction.kt
```diff
@@ -69,6 +75,7 @@ class TransportUpdateManagedIndexMetaDataAction : TransportMasterNodeAction<Upda
     private val log = LogManager.getLogger(javaClass)
     private val client: Client
     private val indexStateManagementHistory: IndexStateManagementHistory
+    private val executor: ManagedIndexMetaDataExecutor = ManagedIndexMetaDataExecutor()
```
jinsoor-amzn, line 78: We can remove `: ManagedIndexMetaDataExecutor` from the variable.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/transport/action/updateindexmetadata/TransportUpdateManagedIndexMetaDataAction.kt
```diff
@@ -87,45 +94,14 @@ class TransportUpdateManagedIndexMetaDataAction : TransportMasterNodeAction<Upda
...
-                    return AcknowledgedResponse(acknowledged)
-                }
+            ManagedIndexMetaDataTask(request.indicesToAddManagedIndexMetaDataTo, request.indicesToRemoveManagedIndexMetaDataFrom),
+            ClusterStateTaskConfig.build(Priority.NORMAL),
+            executor,
+            object : ClusterStateTaskListener {
+                override fun onFailure(source: String, e: Exception) = listener.onFailure(e)
+
+                override fun clusterStateProcessed(source: String, oldState: ClusterState, newState: ClusterState) =
+                    listener.onResponse(AcknowledgedResponse(true))
```
jinsoor-amzn, line 104: What are we acknowledging and is it ok to always return true?

dbbaughe: Acknowledging the cluster state was successfully applied/updated.
And yes, because once clusterStateProcessed is called it means it was successfully processed.

Issue #240, opened 2020-06-16
Failed to create snapshot [BUG]


tocw:
Hello, 
We are running Open Distro release v1.8.0.
The problem I observe is that ISM fails on "create snapshot" action for random indices.
Actually the snapshot is created but ISM fails to complete the action and does not delete the index in the next step.
If I retry failed actionsonly a couple of them at most is able to finish.

Here is the only log message connected to that problem:
```
[2020-06-15T15:00:44,404][INFO ][o.e.s.SnapshotsService   ] [************] snapshot [s3-repository-7-7:offline-retention-test.dev-2020.06.15-15:00:44.361/1GI64gWeRuKuBQ3jHcKFaA] started

[2020-06-15T15:00:44,404][WARN ][o.e.s.SnapshotsService   ] [************] [s3-repository-7-7][offline-retention-test.dev-2020.06.15-15:00:44.361] failed to create snapshot
org.elasticsearch.snapshots.ConcurrentSnapshotExecutionException: [s3-repository-7-7:offline-retention-test.dev-2020.06.15-15:00:44.361]  a snapshot is already running
        at org.elasticsearch.snapshots.SnapshotsService$1.execute(SnapshotsService.java:307) ~[elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.ClusterStateUpdateTask.execute(ClusterStateUpdateTask.java:47) ~[elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.service.MasterService.executeTasks(MasterService.java:702) ~[elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.service.MasterService.calculateTaskOutputs(MasterService.java:324) ~[elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.service.MasterService.runTasks(MasterService.java:219) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.service.MasterService.access$000(MasterService.java:73) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.service.MasterService$Batcher.run(MasterService.java:151) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.cluster.service.TaskBatcher.runIfNotProcessed(TaskBatcher.java:150) [elasticsearch-7.7.0.jar:7.7.0]
[...]

PR #244, opened 2020-06-25
Fixes snapshot bugs

dbbaughe:
*Issue #, if available:*
#240

*Description of changes:*
There were a few bugs found in snapshot from the #240 issue.
1. The snapshot name was not persisting across multiple WaitForSnapshotStep executions as it was added to the info object and not the ActionProperties.
2. The ConcurrentSnapshotExecutionException that was being caught in AttemptSnapshotStep was not working correctly in multi-node setups as the exception is wrapped in a RemoteTransportException
3. There was no try/catch around the WaitForSnapshotStep execute method


dbbaughe:
codecov/path failing because of coverage of the extra branches added in the logic.
Since these are fixing critical bugs in snapshot, will get these in first to make it for the 1.9 release and add additional unit tests after.

PR #254, opened 2020-07-14
Adds support for multi-node run/testing and updates tests

dbbaughe:
*Issue #, if available:*

*Description of changes:*
This adds support in gradle to run and test multinodes.
Updates tests to work in both single/multinode setup.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -156,7 +156,7 @@ testClusters.integTest {
     // When running integration tests it doesn't forward the --debug-jvm to the cluster anymore
     // i.e. we have to use a custom property to flag when we want to debug elasticsearch JVM
     // since we also support multi node integration tests we increase debugPort per node
-    if (System.getProperty("es.debug") != null) {
+    if (System.getProperty("cluster.debug") != null) {
         def debugPort = 5005
         nodes.forEach { node ->
             node.jvmArgs("-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=*:${debugPort}")
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,11 @@ apply plugin: 'elasticsearch.testclusters'
 apply plugin: 'io.gitlab.arturbosch.detekt'
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'org.jetbrains.kotlin.plugin.allopen'
-if (!System.properties.containsKey('tests.rest.cluster') && !System.properties.containsKey('tests.cluster')) {
+
+def usingRemoteCluster = System.properties.containsKey('tests.rest.cluster') || System.properties.containsKey('tests.cluster')
+def usingMultiNode = project.properties.containsKey('numNodes')
+// Only apply jacoco test coverage if we are running a local single node cluster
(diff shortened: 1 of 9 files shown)


src/test/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/resthandler/RestChangePolicyActionIT.kt
```diff
@@ -307,35 +307,39 @@ class RestChangePolicyActionIT : IndexStateManagementRestTestCase() {
...
+                            assertStateEquals(StateMetaData(policy.defaultState, Instant.now().toEpochMilli()), stateMetaDataMap)
+                    )
+                ), explainResponseMap, false)
+        }
 
-        val changePolicy = ChangePolicy(newPolicy.id, null, emptyList(), false)
-        val response = client().makeRequest(RestRequest.Method.POST.toString(),
+        waitFor {
+            val changePolicy = ChangePolicy(newPolicy.id, null, emptyList(), false)
+            val response = client().makeRequest(RestRequest.Method.POST.toString(),
```
bowenlan-amzn, line 327: I am wondering should we write this request inside waitFor block?

dbbaughe: This might be leftover from earlier testing before the other PR with ChangePolicy changes went in.

What was happening was that inside the ChangePolicy RestHandler it was doing a search and getting back an older document with a lower primaryTerm/seqNo which was causing the update to fail and the response coming back was failures: true, failed_indices: [<Something about mismatch between seqNo/PrimaryTerm>]. So this just ensured that the ChangePolicy was correctly applied by retrying if it got a failure because of that race condition. Will rerun a few times to see if thats fixed though.

build.gradle
```diff
@@ -143,21 +147,41 @@ test {
     systemProperty 'tests.security.manager', 'false'
 }
 
+File repo = file("$buildDir/testclusters/repo")
```
bowenlan-amzn, line 150: 👍

build.gradle
```diff
@@ -143,21 +147,41 @@ test {
...
 testClusters.integTest {
     testDistribution = "OSS"
-    if (System.getProperty("cluster.debug") != null) {
-        jvmArgs('-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=5005')
+    // Cluster shrink exception thrown if we try to set numberOfNodes to 1, so only apply if > 1
+    if (_numNodes > 1) numberOfNodes = _numNodes
+    // When running integration tests it doesn't forward the --debug-jvm to the cluster anymore
+    // i.e. we have to use a custom property to flag when we want to debug elasticsearch JVM
+    // since we also support multi node integration tests we increase debugPort per node
+    if (System.getProperty("es.debug") != null) {
```
bowenlan-amzn, line 159: may be helpful to update the README corresponding part 😁

dbbaughe: Accidental change as I was using code from months ago which had a different cluster debug property. Changed back to cluster.debug for cluster JVM.

Issue #266, opened 2020-08-01
[BUG] Force merge fails when taking too long


dbbaughe:
**Describe the bug**
There is currently a bug in force merge which is known and being worked on.

The call to force merge API from the ForceMerge action does not return an acknowledgement, but actually holds the connection open until a rejection is thrown or the force merge completes.

Since force merges can take hours for large indices this causes ISM to go into a weird state for this index as after an hour the lock will expire on the job and it will execute again.

This second execution sees the StepStatus.STARTING from the previous ongoing execution and then decides to disable the job and move it into a StepStatus.FAILED state.

The index will show this state for a while until the first execution potentially completes and then updates the status to StepStatus.COMPLETED and the info to "Starting force merge" while the job is still disabled which is confusing.

Unfortunately the force merge API does not have a way to gracefully close the connection after confirming the force merge task is queued or ongoing so our current idea for a workaround is to let the force merge connection stay open for up to 5 minutes and if it has not thrown a rejection in that 5 minutes (or completed) then just assume it's good to move forward to the WaitFor step. The original call to the force merge API will continue and log the response in a separate coroutine.

dbbaughe:
Fixed in:

https://github.com/opendistro-for-elasticsearch/index-management/pull/267

Issue #250, opened 2020-07-09
Change wording in different actions/steps


dbbaughe:
Some of the steps have messages that might be unclear what they actually mean.

e.g. in transitions when the job is actively evaluating conditions before transitioning to a new state the message says "Attempting to transition" which might be understood as already evaluating to true and actively transitioning to the next state.

This issue is to go over all our messaging and come up with more clear wording to cause less confusion.

dbbaughe:
Action | Step | Possible Messages
------------ | ------------- | -------------
Close | AttemptCloseStep | <ul><li>Successfully closed index</li><li>Failed to close index</li><li>Index had snapshot in progress, retrying closing</li><li>Failed to set index to close</li></ul>
Delete | AttemptDeleteStep | <ul><li>Deleted index</li><li>Failed to delete index</li><li>Index had snapshot in progress, retrying deletion</li><li>Failed to delete index</li></ul>
ForceMerge | AttemptCallForceMergeStep | <ul><li>Started force merge</li><li>Failed to start force merge</li><li>Failed to start force merge</li></ul>
ForceMerge | AttemptSetReadOnlyStep | <ul><li>Set index to read-only</li><li>Failed to set index to read-only</li><li>Failed to set index to read-only</li></ul>
ForceMerge | WaitForForceMergeStep | <ul><li>Force merge completed</li><li>Force merge timed out</li><li>Waiting for force merge to complete</li><li>Unable to retrieve [${ActionProperties.MAX_NUM_SEGMENTS}] from ActionProperties=$actionProperties</li><li>Failed to check segments when waiting for force merge to complete</li><li>Failed to check segments when waiting for force merge to complete</li></ul>
IndexPriority | AttemptSetIndexPriorityStep | <ul><li>Successfully set index priority to $indexPriority</li><li>Failed to set index priority to $indexPriority</li><li>Failed to set index priority to $indexPriority</li></ul>
Notification | AttemptNotificationStep | <ul><li>Successfully sent notification</li><li>Failed to send notification</li></ul>
Open | AttemptOpenStep | <ul><li>Successfully opened index</li><li>Failed to open index: ${managedIndexMetaData.index}</li><li>Failed to set index to open</li></ul>
ReadOnly | SetReadOnlyStep | <ul><li>Set index to read-only</li><li>Failed to set index to read-only</li><li>Failed to set index to read-only</li></ul>
[...]

PR #267, opened 2020-08-03
Fixes force merge failing on long executions, changes some action mes…

dbbaughe:
…saging, adds better try/catch on actions to deal with remote transport exceptions

*Issue #, if available:*
#266, #250

*Description of changes:*
This is a bit of a loaded pull request.
* It changes some of the action message wording to hopefully be a bit more clear. We might change it further to add more information as the "info" object is really supposed to just be a place to dump human friendly/usable information to know what's going on.
* It fixes the force merge bug that would fail if your force merge took too long. This was from the connection not being closed after force merge gets queued/kicked off. We now will hold the connection for up to 5 minutes and then just assume it has succeeded and move on to the wait for step.
* A lot of the action failure messages were useless because of the RemoteTransportException. Correctly catches them in each action and gets the useful error message/cause.
* Added a pre/post execute to steps.
* Fixed some action(s) not correctly catching a snapshot exception that was wrapped in a RemoteTransportException.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
@@ -30,6 +30,7 @@ abstract class Step(val name: String, val managedIndexMetaData: ManagedIndexMeta
         logger.info("Executing $name for ${managedIndexMetaData.index}")
         return this
     }
+
     abstract suspend fun execute(): Step
 
     fun postExecute(logger: Logger): Step {
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.indexstatemanagement.step
 
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.ManagedIndexMetaData
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedindexmetadata.StepMetaData
+import org.apache.logging.log4j.Logger
 import org.elasticsearch.common.io.stream.StreamInput
 import org.elasticsearch.common.io.stream.StreamOutput
 import org.elasticsearch.common.io.stream.Writeable
@@ -25,7 +26,16 @@ import java.util.Locale
 
 abstract class Step(val name: String, val managedIndexMetaData: ManagedIndexMetaData, val isSafeToDisableOn: Boolean = true) {
 
-    abstract suspend fun execute()
+    fun preExecute(logger: Logger): Step {
+        logger.info("Executing $name for ${managedIndexMetaData.index}")
+        return this
+    }
+    abstract suspend fun execute(): Step
+
+    fun postExecute(logger: Logger): Step {
+        logger.info("Finished executing $name for ${managedIndexMetaData.index}")
+        return this
+    }
 
     abstract fun getUpdatedManagedIndexMetaData(currentMetaData: ManagedIndexMetaData): ManagedIndexMetaData
 
@@ -44,9 +54,7 @@ abstract class Step(val name: String, val managedIndexMetaData: ManagedIndexMeta
      */
     abstract fun isIdempotent(): Boolean
 
-    fun getStartingStepMetaData(): StepMetaData {
-        return StepMetaData(name, getStepStartTime().toEpochMilli(), StepStatus.STARTING)
-    }
+    fun getStartingStepMetaData(): StepMetaData = StepMetaData(name, getStepStartTime().toEpochMilli(), StepStatus.STARTING)
 
     fun getStepStartTime(): Instant {
         if (managedIndexMetaData.stepMetaData == null || managedIndexMetaData.stepMetaData.name != this.name) {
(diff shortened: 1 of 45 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/AttemptCallForceMergeStep.kt
```diff
@@ -87,5 +122,9 @@ class AttemptCallForceMergeStep(
 
     companion object {
         const val name = "attempt_call_force_merge"
+        const val FIVE_MINUTES_IN_MILLIS = 1000 * 60 * 5 // how long to wait for the force merge request before moving on
+        const val FIVE_SECONDS_IN_MILLIS = 1000L * 5L // delay
+        fun getFailedMessage(index: String) = "Failed to start force merge [index=$index]"
+        fun getSuccessMessage(index: String) = "Successfully completed force merge [index=$index]"
```
qreshi, line 128: This message can be misleading since this step only calls force merge. `"Successfully completed force merge"` implies the task has completed.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -194,4 +204,15 @@ class AttemptRolloverStep(
             info = info
         )
     }
+
+    companion object {
+        fun getFailedMessage(index: String) = "Failed to rollover index [index=$index]"
+        fun getFailedAliasUpdateMessage(index: String, newIndex: String) =
+            "New index created, but failed to update alias [index=$index, newIndex=$newIndex]"
+        fun getFailedNoValidAliasMessage(index: String) = "Missing rollover_alias index setting [index=$index]"
+        fun getFailedDuplicateRollover(index: String) = "Index has already been rolled over [index=$index]"
```
qreshi, line 213: NP: Can call this `getFailedDuplicateRolloverMessage()` to keep naming scheme consistent.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/AttemptCallForceMergeStep.kt
```diff
@@ -42,35 +51,61 @@ class AttemptCallForceMergeStep(
...
+
+            while (response == null && (Instant.now().toEpochMilli() - startTime) < FIVE_MINUTES_IN_MILLIS) {
+                delay(FIVE_SECONDS_IN_MILLIS)
+                throwable?.let { throw it }
+            }
 
-            // If response is OK then the force merge operation has started
-            if (response.status == RestStatus.OK) {
+            val shadowedResponse = response
+            if (shadowedResponse?.let { it.status == RestStatus.OK } != false) {
```
bowenlan-amzn, line 80: if shadowedResponse is null here, we still update success msg first right? Then it won't override possible failure msg

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
```diff
@@ -25,7 +26,16 @@ import java.util.Locale
...
 abstract class Step(val name: String, val managedIndexMetaData: ManagedIndexMetaData, val isSafeToDisableOn: Boolean = true) {
 
-    abstract suspend fun execute()
+    fun preExecute(logger: Logger): Step {
+        logger.info("Executing $name for ${managedIndexMetaData.index}")
+        return this
+    }
+    abstract suspend fun execute(): Step
+
+    fun postExecute(logger: Logger): Step {
```
bowenlan-amzn, line 36: do we need to return Step here?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/AttemptCallForceMergeStep.kt
dbbaughe: It's half right, if there was a response then it did actually complete before the 5 minute timer. But, if the response is still null it just means we passed 5 minutes and it hasn't completed yet. Updated with another message and a conditional.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/delete/AttemptDeleteStep.kt
```diff
@@ -42,30 +44,51 @@ class AttemptDeleteStep(
...
-            val errorMessage = e.message
-            if (errorMessage != null) mutableInfo["cause"] = errorMessage
-            info = mutableInfo.toMap()
+            resolveException(e)
         }
+
+        return this
+    }
+
+    private fun resolveSnapshotException(e: SnapshotInProgressException) {
```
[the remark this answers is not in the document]
dbbaughe: I have no preference, naming is always hard for me :P

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/Step.kt
dbbaughe: It is with `return this`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/forcemerge/AttemptCallForceMergeStep.kt
dbbaughe: If shadowed response is null it means we hit the 5 minute delay and just assume it's OK to proceed. It now returns a slightly different message saying it successfully called the force merge API whereas if there is a shadowedResponse then it returns a completed message.

Issue #261, opened 2020-07-28
[feature request] Skip rollover on non-write index


jgprogramming:
Hey,

We've noticed a small issue during rollover action. If we rollover an alias manually (for whatever reason) we see that we have multiple managed indices in ISM trying to rollover. This causes alias to rollover (based on time) more often than it should.

Ideally, ISM would either:
* throw an error when rolling over alias due to non-write index ISM policy
* provide configuration to skip rollover if it was already done and managed index is not the write index.

Regards

jgprogramming:
Concrete example of this problem,
![image](https://user-images.githubusercontent.com/20370084/88667578-d9e36980-d0e1-11ea-94d8-38f1b9d4fe29.png)

Previously we've rolled over this index multiple times and let it run for a few hours. Right now each index in hot creates another rollover action.

dbbaughe:
Hi @jgprogramming,

If you have a managed index on the rollover action and you manually rollover the alias I would expect it to fail when it executes rollover as it's not the write alias. Is that not happening?

To be more specific it should fail when it *actually* executes rollover. We moved the condition checking to be done prior to the rollover API call so that we could add our own extra conditions to it in the future. So the current implementation does something like [Evaluate conditions] -[true?]-> [Call Rollover API with no conditions].

We can look into exposing errors that will happen during the rollover API call before evaluating the conditions.

jgprogramming:
Hey @dbbaughe , no, it is not happening. See above screenshot, we have multiple manually rolled over indices that continue to rollover every day or so, based on age condition. 

I suspect it it because rollover step only checks if index was rolled over by ISM (https://github.com/opendistro-for-elasticsearch/index-management/blob/master/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt#L55) and if it was rolled over manually, this flag is not going to be set in managedIndexMetaData.

After that, you rollover on an alias (https://github.com/opendistro-for-elasticsearch/index-management/blob/master/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt#L115) which doesn't throw an error ever, even if index is not a write-index.

I'd suggest to:
1. Create a check after condition check to see if managed index is the write index of the alias
2. Throw an error if managed index is not the write index for the alias
3. Add a boolean parameter to skip rollover instead of throwing error.

[...]

PR #279, opened 2020-08-06
fail rollover on nonwrite index

jgprogramming:
#261

Per above issue discussion, I'm creating a PR that adds a check during rollover action to see if index being rollovered is actually a write index of an alias.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
@@ -24,6 +24,7 @@ import com.amazon.opendistroforelasticsearch.indexstatemanagement.model.managedi
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.step.Step
 import com.amazon.opendistroforelasticsearch.indexstatemanagement.util.evaluateConditions
 import org.apache.logging.log4j.LogManager
+import org.elasticsearch.action.admin.indices.alias.get.GetAliasesRequest
 import org.elasticsearch.action.admin.indices.rollover.RolloverRequest
 import org.elasticsearch.action.admin.indices.rollover.RolloverResponse
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsRequest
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -99,18 +100,28 @@ class AttemptRolloverStep(
...
-            stepStatus = StepStatus.CONDITION_NOT_MET
-            info = mapOf("message" to getAttemptingMessage(indexName), "conditions" to conditions)
+            stepStatus = StepStatus.FAILED
+            info = mapOf("message" to getFailedNonWriteIndexMessage(indexName, alias))
         }
 
         return this
     }
 
+    private fun evaluateWriteIndex(alias: String): Boolean {
```
dbbaughe, line 120: The `aliases.get()` part here will block which we want to avoid. We can refactor this to be a suspend function and utilize the suspendUntil extension function.

This is also a network request which can throw an exception so we'll need to wrap it in a try/catch and handle the exception.

```kotlin
private suspend fun evaluateWriteIndex(alias: String): Boolean {
    try {
        val aliasesRequest = GetAliasesRequest().aliases(alias).indices(indexName)
        val aliasesResponse: GetAliasesResponse = client.admin().indices().suspendUntil { getAliases(aliasesRequest, it) 
        return aliasesResponse.aliases[indexName].firstOrNull()?.writeIndex() == true
    } catch (e: Exception) {
        handleException(e)
    }
}
```

This kind of also opens up the question, if the GetAliases request fails do we really want it to fail the whole Rollover action? Technically this will be called every time the managed index comes to life and checks whether it's time to rollover. So for some people that might rollover after a long time that means it's doing [time it takes to rollover] / [time in job_interval] requests that have the potential to fail for random network reasons. A user can of course set up some retry configuration on the rollover action to reduce permanent failures because of transient network failures, but perhaps having this be a CONDITION_NOT_MET that has a warning message instead? We don't have to tackle this in this PR, just something that perhaps we'll change in the future. For now we can have it treat it as a failure.

And lastly, there is a very slight gap in between the evaluateWriteIndex and rollover call that technically could allow your problem to still happen. Unlikely, but possible. Wondering if in future there is a better way to ensure it can't happen.

solomonsunder: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/skipping-rollover.html

Would it be possible to implement the above way instead? In the case of OpenDistro ILM implementation, maybe set the "rolled_over" : true? This is the response that one gets from the Rollover API anyways.

dbbaughe: Hey @solomonsunder

To be clear you're saying a way for you to manually set rolled_over to be true?

I think in terms of automation there are two choices. If the above happens (write alias is pointing to a different index now) then we can either just accept this index was already rolled over (outside of ISM) and set rolled_over: true and just move on. Or because ISM definitely did not do it we can fail as it might mean there is some outside process doing something unexpectedly. I'm assuming we'll have a mix of wants from different users so perhaps the best choice is to just let the user choose by a rollover setting in the action.

dbbaughe:
Hey @jgprogramming just checking in here. Would you have the time to close out the PR from the comments or would you like us to?

Thanks!

jsirianni:
I have the same problem as this issue, I was happy to see this PR and sad to see that it seems to have become stale. https://github.com/opendistro-for-elasticsearch/index-management/issues/261

dbbaughe:
Closing, will recreate PR separately based off commits in this PR.

PR #285, opened 2020-08-26
Refactors files/classes to prepare for multiple features under the In…

dbbaughe:
…dex Management suite

*Issue #, if available:*

*Description of changes:*


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/elasticapi/ElasticExtensions.kt
```diff
@@ -188,18 +188,17 @@ fun IndexMetadata.shouldDeleteManagedIndexMetaData(): Boolean =
...
-    return this.settings.get(ManagedIndexSettings.POLICY_ID.key)
+    if (ManagedIndexSettings.POLICY_ID.get(this.settings).isNullOrBlank()) return null
+    return ManagedIndexSettings.POLICY_ID.get(this.settings)
 }
 
 /**
  * Returns the current rollover_alias if it exists otherwise returns null.
  * */
 fun IndexMetadata.getRolloverAlias(): String? {
-    if (this.settings.get(ManagedIndexSettings.ROLLOVER_ALIAS.key).isNullOrBlank()) return null
```
bowenlan-amzn, line 200: I feel this way is more straightforward

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/IndexStateManagementHistory.kt
```diff
@@ -177,20 +180,20 @@ class IndexStateManagementHistory(
     }
 
     @Suppress("NestedBlockDepth")
-    suspend fun addHistory(managedIndexMetaData: List<ManagedIndexMetaData>) {
+    suspend fun addManagedIndexMetaDataHistory(managedIndexMetaData: List<ManagedIndexMetaData>) {
```
bowenlan-amzn, line 183: can we use Metadata 😀

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/IndexManagementPlugin.kt
```diff
@@ -59,32 +62,27 @@ import org.elasticsearch.threadpool.ThreadPool
...
     companion object {
-        const val PLUGIN_NAME = "opendistro-ism"
-        const val ISM_BASE_URI = "/_opendistro/_ism"
+        const val PLUGIN_NAME = "opendistro-im"
+        const val OPEN_DISTRO_BASE_URI = "/_opendistro"
+        const val ISM_BASE_URI = "$OPEN_DISTRO_BASE_URI/_ism"
         const val POLICY_BASE_URI = "$ISM_BASE_URI/policies"
-        const val INDEX_STATE_MANAGEMENT_INDEX = ".opendistro-ism-config"
-        const val INDEX_STATE_MANAGEMENT_JOB_TYPE = "opendistro-managed-index"
+        const val INDEX_MANAGEMENT_INDEX = ".opendistro-ism-config"
```
bowenlan-amzn, line 76: Not sure why change this to INDEX_MANAGEMENT_INDEX, it should still for state management right

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/IndexManagementPlugin.kt
```diff
@@ -59,32 +62,27 @@ import org.elasticsearch.threadpool.ThreadPool
...
+        const val INDEX_MANAGEMENT_JOB_TYPE = "opendistro-index-management"
```
dbbaughe, line 77: As a note: Job Scheduler only lets us specify one job type which is why we changed this to be more generic for the whole plugin. At the same time it does not actually use this so it's not that important.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/IndexManagementIndices.kt
```diff
@@ -36,17 +36,17 @@ import org.elasticsearch.common.settings.Settings
...
 @OpenForTesting
-class IndexStateManagementIndices(
+class IndexManagementIndices(
     private val client: IndicesAdminClient,
     private val clusterService: ClusterService
 ) {
 
     private val logger = LogManager.getLogger(javaClass)
 
     fun checkAndUpdateISMConfigIndex(actionListener: ActionListener<AcknowledgedResponse>) {
```
bowenlan-amzn, line 46: do we want to change ISM to IM in the function name

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/IndexManagementPlugin.kt
dbbaughe: We are adding more types of index management jobs into this config index; Rollups and Transforms. So this is not just for ISM anymore. Unfortunately we cannot change the index name itself in `.opendistro-**ism**-config` so this will just have to be a legacy artifact, but we can rename the variable at least.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/elasticapi/ElasticExtensions.kt
dbbaughe: Can change it back, was originally changed because I had this setting deprecated with a new setting name and this old way would not be able to default back to deprecated value.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/IndexStateManagementHistory.kt
dbbaughe: Will keep it same for now as it's following parameters convention

PR #290, opened 2020-09-09
Adds support for dynamically updatable search analyzers

setiah:
Issue #, if available:

Description of changes:
This change is to add support for hot update of search analyzers in Elasticsearch. It exposes a new REST Api `POST /_opendistro/_analyzer/refresh_search_analyzer/{index}`, which when invoked, refreshes those analyzers in memory which were created with is_updateable flag as true. This can be used to refresh synonym list used for search time analysis, without closing and reopening the index.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
@@ -15,13 +15,15 @@
 
 package com.amazon.opendistroforelasticsearch.indexmanagement.refreshanalyzer
 
-import org.elasticsearch.action.support.broadcast.BroadcastRequest
-import org.elasticsearch.common.io.stream.StreamInput
-import java.io.IOException
+import org.elasticsearch.action.ActionType
+import org.elasticsearch.common.io.stream.Writeable
 
-class RefreshSynonymAnalyzerRequest : BroadcastRequest<RefreshSynonymAnalyzerRequest> {
-    constructor(vararg indices: String) : super(*indices)
+class RefreshSearchAnalyzerAction : ActionType<RefreshSearchAnalyzerResponse>(NAME, reader) {
+    companion object {
+        const val NAME = "indices:admin/refresh_search_analyzer"
+        val INSTANCE = RefreshSearchAnalyzerAction()
+        val reader = Writeable.Reader { inp -> RefreshSearchAnalyzerResponse(inp) }
+    }
 
-    @Throws(IOException::class)
-    constructor(inp: StreamInput) : super(inp)
+    override fun getResponseReader(): Writeable.Reader<RefreshSearchAnalyzerResponse> = reader
 }
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
@@ -20,7 +20,7 @@ import org.elasticsearch.common.io.stream.Writeable
 
 class RefreshSearchAnalyzerAction : ActionType<RefreshSearchAnalyzerResponse>(NAME, reader) {
     companion object {
-        const val NAME = "indices:admin/refresh_search_analyzer"
+        const val NAME = "indices:admin/_refresh_search_analyzers"
         val INSTANCE = RefreshSearchAnalyzerAction()
         val reader = Writeable.Reader { inp -> RefreshSearchAnalyzerResponse(inp) }
     }
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
@@ -20,7 +20,7 @@ import org.elasticsearch.common.io.stream.Writeable
 
 class RefreshSearchAnalyzerAction : ActionType<RefreshSearchAnalyzerResponse>(NAME, reader) {
     companion object {
-        const val NAME = "indices:admin/_refresh_search_analyzers"
+        const val NAME = "indices:admin/refresh_search_analyzers"
         val INSTANCE = RefreshSearchAnalyzerAction()
         val reader = Writeable.Reader { inp -> RefreshSearchAnalyzerResponse(inp) }
     }
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.indexmanagement.refreshanalyzer
 
 import org.apache.logging.log4j.LogManager
 import org.elasticsearch.action.support.DefaultShardOperationFailedException
+import org.elasticsearch.action.support.DefaultShardOperationFailedException.readShardOperationFailed
 import org.elasticsearch.action.support.broadcast.BroadcastResponse
 import org.elasticsearch.common.io.stream.StreamInput
 import org.elasticsearch.common.io.stream.StreamOutput
@@ -29,22 +30,21 @@ import java.util.function.Function
 
 class RefreshSearchAnalyzerResponse : BroadcastResponse {
 
-    private var results: MutableMap<String, List<String>> = HashMap()
-    private var shardFailures: MutableList<FailedShardDetails> = mutableListOf()
-    private var temp: List<DefaultShardOperationFailedException> = mutableListOf()
-
     protected var logger = LogManager.getLogger(javaClass)
+
+    private lateinit var shardResponses: MutableList<RefreshSearchAnalyzerShardResponse>
+    private lateinit var shardFailures: MutableList<DefaultShardOperationFailedException>
 
     @Throws(IOException::class)
     constructor(inp: StreamInput) : super(inp) {
         val resultSize: Int = inp.readVInt()
         for (i in 0..resultSize) {
-            results.put(inp.readString(), inp.readStringArray().toList())
+            shardResponses.add(RefreshSearchAnalyzerShardResponse(inp))
         }
 
         val failureSize: Int = inp.readVInt()
         for (i in 0..failureSize) {
-            shardFailures.add(FailedShardDetails(inp.readString(), inp.readInt(), inp.readString()))
+            shardFailures.add(readShardOperationFailed(inp))
         }
     }
 
@@ -53,58 +53,54 @@ class RefreshSearchAnalyzerResponse : BroadcastResponse {
         successfulShards: Int,
         failedShards: Int,
         shardFailures: List<DefaultShardOperationFailedException>,
-        results: MutableMap<String, List<String>>
+        shardResponses: List<RefreshSearchAnalyzerShardResponse>
     ) : super(
             totalShards, successfulShards, failedShards, shardFailures
     ) {
-        this.results = results
-        this.temp = shardFailures
-        for (failure in shardFailures) {
-            this.shardFailures.add(FailedShardDetails(failure.index()!!, failure.shardId(), failure.reason()))
-        }
+        this.shardResponses = shardResponses.toMutableList()
+        this.shardFailures = shardFailures.toMutableList()
     }
 
     @Throws(IOException::class)
     override fun toXContent(builder: XContentBuilder, params: Params?): XContentBuilder? {
         builder.startObject()
-        RestActions.buildBroadcastShardsHeader(builder, params, totalShards, successfulShards, -1, failedShards, null)
-
-        builder.startArray("_successful")
-        for (index in results.keys) {
-            builder.startObject()
-            val reloadedAnalyzers: List<String> = results.get(index)!!
-            builder.field("index", index)
-            builder.startArray("refreshed_analyzers")
+        RestActions.buildBroadcastShardsHeader(builder, params, totalShards, successfulShards, -1, failedShards, shardFailures.toTypedArray())
+        builder.startArray("successful_refreshes")
+        val successfulIndices = getSuccessfulRefreshDetails()
+        for (index in successfulIndices.keys) {
+            val reloadedAnalyzers = successfulIndices.get(index)!!
+            builder.startObject().field("index", index).startArray("refreshed_analyzers")
             for (analyzer in reloadedAnalyzers) {
                 builder.value(analyzer)
             }
-            builder.endArray()
-            builder.endObject()
+            builder.endArray().endObject()
         }
-        builder.endArray()
-
-        builder.startArray("_failed")
-        for (failure in shardFailures) {
-            builder.startObject()
-            builder.field("index", failure.index)
-            builder.field("shardId", failure.shardId)
-            builder.field("failureReason", failure.failureReason)
-            builder.endObject()
-        }
-        builder.endArray()
-
-        builder.endObject()
+        builder.endArray().endObject()
         return builder
     }
 
+    // TODO: restrict it for testing
+    fun getSuccessfulRefreshDetails(): MutableMap<String, List<String>> {
+        var successfulRefreshDetails: MutableMap<String, List<String>> = HashMap()
+        var failedIndices = mutableSetOf<String>()
+        for (failure in shardFailures) {
+            failedIndices.add(failure.index()!!)
+        }
+        for (response in shardResponses) {
+            if (!failedIndices.contains(response.index)) {
+                successfulRefreshDetails.putIfAbsent(response.index, response.reloadedAnalyzers)
+            }
+        }
+        return successfulRefreshDetails
+    }
+
     companion object {
-        private val PARSER = ConstructingObjectParser<RefreshSearchAnalyzerResponse, Void>("refresh_search_analyzer", true,
+        private val PARSER = ConstructingObjectParser<RefreshSearchAnalyzerResponse, Void>("_refresh_search_analyzers", true,
                 Function { arg: Array<Any> ->
                     val response = arg[0] as RefreshSearchAnalyzerResponse
                     RefreshSearchAnalyzerResponse(response.totalShards, response.successfulShards, response.failedShards,
-                            response.temp, response.results)
+                            response.shardFailures, response.shardResponses)
                 })
-
         init {
             declareBroadcastFields(PARSER)
         }
@@ -114,23 +110,14 @@ class RefreshSearchAnalyzerResponse : BroadcastResponse {
     override fun writeTo(out: StreamOutput) {
         super.writeTo(out)
 
-        out.writeVInt(results.size)
-        for ((key, value) in results.entries) {
-            out.writeString(key)
-            out.writeStringArray(value.toTypedArray())
+        out.writeVInt(shardResponses.size)
+        for (response in shardResponses) {
+            response.writeTo(out)
         }
 
         out.writeVInt(shardFailures.size)
         for (failure in shardFailures) {
-            out.writeString(failure.index)
-            out.writeInt(failure.shardId)
-            out.writeString(failure.failureReason)
+            failure.writeTo(out)
         }
-    }
-
-    class FailedShardDetails(index: String, shardId: Int, failureReason: String) {
-        val index = index
-        val shardId = shardId
-        val failureReason = failureReason
     }
 }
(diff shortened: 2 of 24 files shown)


setiah:
In the process of adding more UTs to close the codecov/patch gap

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
```diff
@@ -0,0 +1,136 @@
...
+        successfulShards: Int,
+        failedShards: Int,
+        shardFailures: List<DefaultShardOperationFailedException>,
+        results: MutableMap<String, List<String>>
+    ) : super(
+            totalShards, successfulShards, failedShards, shardFailures
+    ) {
+        this.results = results
+        this.temp = shardFailures
+        for (failure in shardFailures) {
```
qreshi, line 62: Can replace this with a `map`

Ex.
```
this.shardFailures = shardFailures.map{ FailedShardDetails(it.index()!!, it.shardId(), it.reason()) }
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
```diff
@@ -0,0 +1,136 @@
...
+    private var results: MutableMap<String, List<String>> = HashMap()
+    private var shardFailures: MutableList<FailedShardDetails> = mutableListOf()
+    private var temp: List<DefaultShardOperationFailedException> = mutableListOf()
+
+    protected var logger = LogManager.getLogger(javaClass)
+
+    @Throws(IOException::class)
+    constructor(inp: StreamInput) : super(inp) {
+        val resultSize: Int = inp.readVInt()
+        for (i in 0..resultSize) {
```
dbbaughe, line 41: Is this a common pattern you saw elsewhere for reading off streaminput?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
```diff
@@ -0,0 +1,136 @@
...
+            results.put(inp.readString(), inp.readStringArray().toList())
+        }
+
+        val failureSize: Int = inp.readVInt()
+        for (i in 0..failureSize) {
+            shardFailures.add(FailedShardDetails(inp.readString(), inp.readInt(), inp.readString()))
```
dbbaughe, line 47: nit: Make FailedShardDetails writeable and just pass in the streaminput/output

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
```diff
@@ -0,0 +1,136 @@
...
+                builder.value(analyzer)
+            }
+            builder.endArray()
+            builder.endObject()
+        }
+        builder.endArray()
+
+        builder.startArray("_failed")
+        for (failure in shardFailures) {
+            builder.startObject()
```
qreshi, line 88: Can chain these `builder` calls to make it look a little cleaner:

```
builder.startObject()
    .field("index", failure.index)
    .field("shardId", failure.shardId)
    .field("failureReason", failure.failureReason)
    .endObject()
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerShardResponse.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.elasticsearch.common.io.stream.StreamInput
+import org.elasticsearch.common.io.stream.StreamOutput
+import org.elasticsearch.index.shard.ShardId
+import java.io.IOException
+
+class RefreshSearchAnalyzerShardResponse : BroadcastShardResponse {
+    var indexName: String
+    var reloadedAnalyzers: List<String>
+
+    constructor(`in`: StreamInput) : super(`in`) {
```
qreshi, line 28: Can replace `in` with something like `si`. I think `si` is used elsewhere in the code as well and you can avoid using a keyword.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
```diff
@@ -0,0 +1,117 @@
...
+    ) {
+        this.analysisRegistry = analysisRegistry
+        this.indicesService = indicesService
+    }
+
+    private val indicesService: IndicesService
+    private val analysisRegistry: AnalysisRegistry
+
+    @Throws(IOException::class)
+    override fun readShardResult(`in`: StreamInput): RefreshSearchAnalyzerShardResponse? {
```
qreshi, line 69: Same here, can use `si`

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
```diff
@@ -0,0 +1,270 @@
...
+        val indexName = "testindex"
+
+        for (i in 0 until numNodes) {
+            writeToFile("$buildDir/testclusters/integTest-$i/config/pacman_synonyms.txt", "hello, hola")
+        }
+
+        val settings: Settings = Settings.builder()
+                .loadFromSource(getSearchAnalyzerSettings(), XContentType.JSON)
+                .build()
+        // val mappings: String = "\"properties\":{\"title\":{\"type\": \"text\",\"analyzer\" : \"standard\",\"search_analyzer\": \"my_synonyms\"}}"
```
qreshi, line 90: NP: Can remove this if not needed

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
dbbaughe, line 69: Remove nullable return type as it seems to never be null

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
```diff
@@ -0,0 +1,117 @@
...
+        return RefreshSearchAnalyzerShardResponse(`in`)
+    }
+
+    override fun newResponse(
+        request: RefreshSearchAnalyzerRequest?,
```
dbbaughe, line 74: Just checking: can this be null or was this just from autocomplete from IDE

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
```diff
@@ -0,0 +1,270 @@
...
+
+    companion object {
+
+        fun writeToFile(filePath: String, contents: String) {
+            var path = org.elasticsearch.common.io.PathUtils.get(filePath)
+            Files.newBufferedWriter(path, Charset.forName("UTF-8")).use { writer -> writer.write(contents) }
+        }
+
+        fun deleteFile(filePath: String) {
+            // org.elasticsearch.common.io.PathUtils.get(filePath)
```
qreshi, line 184: Can remove this as well if not needed

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
```diff
@@ -0,0 +1,117 @@
...
+        totalShards: Int,
+        successfulShards: Int,
+        failedShards: Int,
+        results: List<RefreshSearchAnalyzerShardResponse>,
+        shardFailures: List<DefaultShardOperationFailedException>,
+        clusterState: ClusterState?
```
dbbaughe, line 80: Just checking: can this be null or was this just from autocomplete from IDE

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
```diff
@@ -0,0 +1,117 @@
...
+    ): RefreshSearchAnalyzerResponse {
+        val shardResponses: MutableMap<String, List<String>> = HashMap()
+        for (response in results) {
+            shardResponses.put(response.indexName, response.reloadedAnalyzers)
+        }
+        return RefreshSearchAnalyzerResponse(totalShards, successfulShards, failedShards, shardFailures, shardResponses)
+    }
+
+    @Throws(IOException::class)
+    override fun readRequestFrom(`in`: StreamInput): RefreshSearchAnalyzerRequest? {
```
dbbaughe, line 90: Remove nullable

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RestRefreshSearchAnalyzerAction.kt
```diff
@@ -0,0 +1,60 @@
...
+        val refreshSearchAnalyzerRequest: RefreshSearchAnalyzerRequest = RefreshSearchAnalyzerRequest()
+                .indices(*indices)
+
+        return RestChannelConsumer { channel ->
+            client.execute(RefreshSearchAnalyzerAction.INSTANCE, refreshSearchAnalyzerRequest, RestToXContentListener(channel))
+        }
+    }
+
+    companion object {
+        const val REFRESH_SYNONYM_ANALYZER_URI = "$ANALYZER_BASE_URI/refresh_search_analyzer/{index}"
```
dbbaughe, line 58: Do you want this to be `/_opendistro/_analyzer/refresh_search_analyzer/{index}`?
Are we expecting to add a lot more APIs under _analyzer?
What about just `/{index}/_refresh_search_analyzer`

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RestRefreshSearchAnalyzerActionIT.kt
```diff
@@ -0,0 +1,64 @@
...
+            val expectedErrorMessage = mapOf(
+                    "error" to mapOf(
+                            "root_cause" to listOf<Map<String, Any>>(
+                                    mapOf("type" to "parse_exception", "reason" to "request body is required")
+                            ),
+                            "type" to "parse_exception",
+                            "reason" to "request body is required"
+                    ),
+                    "status" to 400
+            )
```
dbbaughe, line 44: This seems to be not what you are trying to test? If testing for missing indices I'd expect to see an expected message for `throw IllegalArgumentException("Missing indices")`

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
```diff
@@ -0,0 +1,270 @@
...
+        for (i in 0 until numNodes) {
+            writeToFile("$buildDir/testclusters/integTest-$i/config/pacman_synonyms.txt", "hello, hola")
+        }
+
+        val settings: Settings = Settings.builder()
+                .loadFromSource(getIndexAnalyzerSettings(), XContentType.JSON)
+                .build()
+        createIndex(indexName, settings, getAnalyzerMapping())
+        ingestData(indexName)
+        Thread.sleep(1000) // wait for refresh_interval
```
dbbaughe, line 44: Can we remove this and just wrap the next assertion in a waitFor block

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
```diff
@@ -0,0 +1,270 @@
...
+            writeToFile("$buildDir/testclusters/integTest-$i/config/pacman_synonyms.txt", "hello, hola")
+        }
+
+        val settings: Settings = Settings.builder()
+                .loadFromSource(getSearchAnalyzerSettings(), XContentType.JSON)
+                .build()
+        // val mappings: String = "\"properties\":{\"title\":{\"type\": \"text\",\"analyzer\" : \"standard\",\"search_analyzer\": \"my_synonyms\"}}"
+        createIndex(indexName, settings, getAnalyzerMapping())
+        ingestData(indexName)
+        Thread.sleep(1000) // wait for refresh_interval
```
dbbaughe, line 93: Same as above -> use waitFor

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RestRefreshSearchAnalyzerAction.kt
setiah: It looked liked the latter one earlier but i changed it to keep up with the odfe naming convention. I can change it back to `/{index}/_refresh_search_analyzer` if you feel this is ok.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
setiah: sure

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
setiah: sure, will do

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
setiah: sure

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RestRefreshSearchAnalyzerActionIT.kt
setiah: the uri change caused the exception behavior to change. I will revert back the uri and change this back

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RestRefreshSearchAnalyzerAction.kt
dbbaughe: That would be my preference, the other APIs look like that because they are managing resources under ODFE mostly. This is an index level operation so feel like it should be fine to support in the format elastic uses. In that thought maybe it's not a great idea, not sure what happens if we add this and then they add a _refresh_search_analyzer too on the off chance. So maybe it's good to keep it under opendistro. I'm impartial to either or.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
setiah: Good point. I wanted to do some refactoring on this but couldn't get to it earlier. It's done now, the new version does not use this class now.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
setiah: I did some refactoring, this won't be needed now.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerShardResponse.kt
setiah: sure, done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
setiah: done

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
setiah: sure, this code is removed however.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
setiah: sure

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
setiah: sure

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
qreshi: Also, does this constructor only take the streaminput in the format outputted by `writeTo`? if so, you can change `writeTo` to do something like:

```
out.writeList(shardResponses)
```

and then here you can read the list and pass a class reference to the element of the list (assuming it's `Writeable`)

```
inp.readList(::RefreshSearchAnalyzerShardResponse)
```

You can find a similar example of this [in the Monitor data class for Alerting](https://github.com/opendistro-for-elasticsearch/alerting/blob/7320bc6071804f810d4245dda701ebebc9cbeb3c/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/Monitor.kt#L88)

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
```diff
@@ -20,7 +20,7 @@ import org.elasticsearch.common.io.stream.Writeable
 
 class RefreshSearchAnalyzerAction : ActionType<RefreshSearchAnalyzerResponse>(NAME, reader) {
     companion object {
-        const val NAME = "indices:admin/refresh_search_analyzer"
+        const val NAME = "indices:admin/_refresh_search_analyzers"
```
dbbaughe, line 23: Is it common to prefix _ on the action names?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerResponse.kt
setiah: I found this pattern in [BroadcastResponse.java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/support/broadcast/BroadcastResponse.java#L76-L81), although the one that Mohammad suggested looks to be a better way. I have modified it in my latest commit. Thanks for the suggestion Mo!

dbbaughe (dismissed):

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerAction.kt
setiah: You're right, it's not. Will change back to `indices:admin/refresh_search_analyzers`

qreshi (approved):

bowenlan-amzn (approved):

PR #309, opened 2020-10-13
Change release workflow to use new staging bucket for artifacts

camerski:
PLEASE DO NOT MERGE THIS REQUEST UNTIL ASKED. We need to coordinate the merge with updating Github secrets.

*Description of changes:*
The infrastructure team is separating the production and staging locations into different AWS accounts. Plugins need to modify their workflows to publish to the new locations.

This PR changes the CD workflow to add a build number and write the zip, deb, and rpm plugin artifacts to staging.artifacts.opendistroforelasticsearch.amazon.com. The write to S3 currently fails (see https://github.com/camerski/index-management/actions/runs/305187372) because the secrets have not been updated; the secrets will be updated at the same time this PR is merged.

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -50,11 +50,25 @@ jobs:
       # This step uses the upload-artifact Github action: https://github.com/actions/upload-artifact
       - name: Upload Artifacts to S3
         run: |
-          s3_path=s3://artifacts.opendistroforelasticsearch.amazon.com/downloads
-          aws s3 cp index-management-artifacts/*.zip $s3_path/elasticsearch-plugins/opendistro-index-management/
-          aws s3 cp index-management-artifacts/*.deb $s3_path/debs/opendistro-index-management/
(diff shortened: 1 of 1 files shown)


dbbaughe:
Seems like the build number being used is just the arbitrary `GITHUB_RUN_NUMBER` which is increased on every run. Is that what we want instead of using our versioning as the build number?

dbbaughe (approved):

dbbaughe:
> > Seems like the build number being used is just the arbitrary `GITHUB_RUN_NUMBER` which is increased on every run. Is that what we want instead of using our versioning as the build number?
> 
> That's correct. We want to ensure that each run generates a unique artifact and does not overwrite old builds. The plugin version is still part of the file name. If there's a better way to generate unique build numbers then I'm all ears :)

Nope, just curious 😛

bowenlan-amzn (approved):

gaiksaya:
Branch is out-of-date and cannot sync with recent and more changes due to access issues. Hence closing this PR and raising a new one https://github.com/opendistro-for-elasticsearch/index-management/pull/378

PR #315, opened 2020-10-21
skip RefreshSearchAnalyzerActionIT for Infra IT

bowenlan-amzn:
*Issue #, if available:*

*Description of changes:*
since these 3 tests require to write to Elasticsearch config path, which in local `./gradlew integTest` is `$buildDir/testclusters/integTest-$i/config/`, but when running against a remote cluster like [this](https://github.com/opendistro-for-elasticsearch/opendistro-build/actions/runs/318538503/workflow#L212)
`./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername=es-integrationtest`, since config path change, test would fail

by checking clustername to set a flag to skip tests


camerski (approved):

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/refreshanalyzer/RefreshSearchAnalyzerActionIT.kt
```diff
@@ -21,13 +21,15 @@ import org.elasticsearch.client.Request
...
 import org.elasticsearch.common.xcontent.XContentType
+import org.junit.Assume
 import java.io.InputStreamReader
 import java.nio.charset.Charset
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
 
 class RefreshSearchAnalyzerActionIT : IndexManagementRestTestCase() {
     fun `test index time analyzer`() {
+        Assume.assumeTrue(isLocalTest)
```
setiah, line 32: A cleaner approach - This check can be added to a common function, which can then run before every test in this file with `@Before`

```
  @Before
  public void checkIfLocalCluster() {
    Assume.assumeTrue(isLocalTest)
  }
```

setiah (approved):

camerski (approved):

PR #316, opened 2020-10-22
Change integration tests to use relative paths for snapshot repositories

camerski:
*Description of changes:*
Change the integration tests to use relative paths for snapshot repositories. This breaks the dependency on having the path match the value in `path.repo`.

*Testing done:*
`gradlew integTest` succeeded


bowenlan-amzn (approved):

dbbaughe:
If I remember that change was to fix it previously breaking for infra tests, not sure how it went from fixing it to now being the reason it's breaking 🤔

dbbaughe (approved):

PR #337, opened 2020-11-17
Adds rollup metadata service that deals with initializing and crud operations

dbbaughe:
…erations for the a rollup metadata

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupMetadataService.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupMetadataService.kt
@@ -15,12 +15,407 @@
 
 package com.amazon.opendistroforelasticsearch.indexmanagement.rollup
 
+import com.amazon.opendistroforelasticsearch.indexmanagement.IndexManagementPlugin
+import com.amazon.opendistroforelasticsearch.indexmanagement.elasticapi.parseWithType
+import com.amazon.opendistroforelasticsearch.indexmanagement.elasticapi.suspendUntil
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupMetadataService.kt
```diff
@@ -15,12 +15,407 @@
...
+    }
+
+    /**
+     * Get the rounding strategy for the given time interval in the DateHistogram.
+     * This is used to calculate time windows by rounding the given time based on the interval.
+     */
+    // TODO: Could make this an extension function of DateHistogram and add to some utility file
+    private fun getRoundingStrategy(dateHistogram: DateHistogram): Rounding {
+        val intervalString = (dateHistogram.calendarInterval ?: dateHistogram.fixedInterval) as String
+        // TODO: Make sure the interval string is validated before getting here so we don't get errors
```
qreshi, line 242: Let's double check this to confirm when is the earliest point in the code that we catch an incorrect DateHistogram interval string. If it isn't caught before here, we would enter the `else` statement and may not handle the case correctly.

Perhaps we could leverage the DateHistogram builder and try to enter the given interval string and catch any exceptions that are thrown. If this approach is valid, we could even put it in the `init` of `Rollup` to validate early.

dbbaughe: Will cover TODOs and any cleanups needed in follow up PRs to unblock the PRs that depend on these.

qreshi (approved):

thalurur (approved):

PR #341, opened 2020-11-18
Adds rollup runner and integration tests

dbbaughe:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
@@ -15,20 +15,41 @@
 
 package com.amazon.opendistroforelasticsearch.indexmanagement.rollup
 
+import com.amazon.opendistroforelasticsearch.indexmanagement.elasticapi.retry
+import com.amazon.opendistroforelasticsearch.indexmanagement.elasticapi.suspendUntil
+import com.amazon.opendistroforelasticsearch.indexmanagement.rollup.action.get.GetRollupAction
+import com.amazon.opendistroforelasticsearch.indexmanagement.rollup.action.get.GetRollupRequest
+import com.amazon.opendistroforelasticsearch.indexmanagement.rollup.action.get.GetRollupResponse
+import com.amazon.opendistroforelasticsearch.indexmanagement.rollup.action.index.IndexRollupAction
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
```diff
@@ -109,5 +130,331 @@ object RollupRunner : ScheduledJobRunner,
...
+
+    private suspend fun releaseLockForRollupJob(context: JobExecutionContext, lock: LockModel): Boolean {
+        return context.lockService.suspendUntil { release(lock, it) }
+    }
+
+    // TODO: Clean up runner
+    // TODO: Scenario: The rollup job is finished, but I (the user) want to redo it all again
+    /*
+    * TODO situations:
+    *  There is a rollup.metadataID and doc but theres no job in target index?
```
qreshi, line 212: Just a heads up, I believe we do recover from this scenario now, however, it's not exactly as the TODO message expects:

* During `isJobValid`, we'll validate source and target index
* If we get the condition in this scenario, where rollup.metadataID and doc exist but there is no job in target index, we resolve it by updating the mappings for the target index and add the job in

So we recover from it but neither start over or move to `FAILED`, I suppose it would recover silently and continue from wherever the metadata has left off.

We can go over this behavior as part of reviewing retry logic later on to discuss if it's what we want or not.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
```diff
@@ -109,5 +130,331 @@ object RollupRunner : ScheduledJobRunner,
...
+                            // If we fail to renew the lock it doesn't mean we need to perm fail the job, we can just return early
+                            // and let the next execution try to process the data from where this one left off
+                            return
+                        }
+                    } catch (e: RollupMetadataException) {
+                        // Rethrow this exception so it doesn't get consumed here
+                        logger.info("RollupMetadataException being thrown", e)
+                        throw e
+                    } catch (e: Exception) {
+                        // TODO: Should update metadata and disable job here instead of allowing the rollup to keep going
```
qreshi, line 322: This TODO should also be a candidate for inclusion whenever we work on retry logic.

qreshi (approved):

qreshi (approved):

thalurur (approved):

PR #371, opened 2021-01-06
Introducing new ISM Action to create rollups

thalurur:
*Issue #, if available:*
N/A

*Description of changes:*
A new ISM Action called RollupAction is introduced to create single run rollups


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/action/RollupAction.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.amazon.opendistroforelasticsearch.indexmanagement.indexstatemanagement.step.rollup.AttemptCreateRollupJobStep
+import com.amazon.opendistroforelasticsearch.indexmanagement.indexstatemanagement.step.rollup.WaitForRollupCompletionStep
+import com.amazon.opendistroforelasticsearch.indexmanagement.rollup.model.ISMRollup
+import org.elasticsearch.client.Client
+import org.elasticsearch.cluster.service.ClusterService
+
+class RollupAction(
+    clusterService: ClusterService,
+    client: Client,
+    ismRollup: ISMRollup,
```
dbbaughe, line 31: Can probably remove this as it's on the RollupActionConfig that is already being passed in.

thalurur: right, agree. Will remove

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/action/RollupActionConfig.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    @Throws(IOException::class)
+    override fun writeTo(out: StreamOutput) {
+        super.writeTo(out)
+        ismRollup.writeTo(out)
+        out.writeInt(actionIndex)
+    }
+
+    companion object {
+        const val ISM_ROLLUP_FIELD = "ism_rollup"
```
dbbaughe, line 69: Do you think we need the ism prefix?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/managedindexmetadata/ActionProperties.kt
```diff
@@ -28,17 +28,23 @@ import org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken
 /** Properties that will persist across steps of a single Action. Will be stored in the [ActionMetaData]. */
 data class ActionProperties(
     val maxNumSegments: Int? = null,
-    val snapshotName: String? = null
+    val snapshotName: String? = null,
+    val rollupId: String? = null,
+    val hasRollupFailed: Boolean? = null
```
dbbaughe, line 33: Possible to namespace these under "rollup" since they are related? Admittedly this was just a dumping ground of random values that needed to persist across step executions, but worried it'll continue to just grow and get confusing to reason about, especially if some properties are tied together.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/action/RollupActionConfig.kt
thalurur: kept it cause the schema is specific to ism rollup action, but not particular about it. However, the action is also called rollup so the json will look like below, if we remove

`{"rollup": {"rollup": { .. }}}` instead of `{"rollup": {"ism_rollup": { .. }}}`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/ISMRollup.kt
```diff
@@ -0,0 +1,199 @@
...
+data class ISMRollup(
+    val description: String,
+    val targetIndex: String,
+    val pageSize: Int,
+    val dimensions: List<Dimension>,
+    val metrics: List<RollupMetrics>
+) : ToXContentObject, Writeable {
+
+    init {
+        require(pageSize in Rollup.MINIMUM_PAGE_SIZE..Rollup.MAXIMUM_PAGE_SIZE) { "Page size must be between ${Rollup.MINIMUM_PAGE_SIZE} " +
```
dbbaughe, line 47: nit: Would be good to somehow remove the duplication between the validations here and the other rollup, seems like if we add any in Rollup that applies here we have to remember to do it here too. Can be a TODO for a later time though.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/ISMRollup.kt
```diff
@@ -0,0 +1,199 @@
...
+                                Dimension.Type.HISTOGRAM -> Histogram(sin)
+                            }
+                    )
+                }
+                dimensionsList.toList()
+            },
+            metrics = sin.readList(::RollupMetrics)
+    )
+
+    override fun toString(): String {
```
dbbaughe, line 117: I believe the doc id has a limit which this could hit?

dbbaughe: Ah I see the DigestUtils above, ignore this

src/main/resources/mappings/opendistro-ism-history.json
```diff
@@ -93,6 +93,17 @@
                       "type": "keyword"
                     }
                   }
+                },
+                "rollup_id": {
+                  "type": "text",
+                  "fields": {
+                    "raw": {
```
dbbaughe, line 100: Let's change to "keyword", not sure when "raw" got introduced but we have a mix of "raw" and "keyword" now for the sub field type keyword and majority is keyword so we'll try to stick with that.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/ISMRollup.kt
thalurur: I am doing a fixed length hash - sha1

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/managedindexmetadata/ActionProperties.kt
thalurur: will add a TODO for this, agree this can blow up if we add more fields to this

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/ISMRollup.kt
thalurur: sure will add a TODO

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
```diff
@@ -0,0 +1,123 @@
...
+
+        // Creating a rollup job
+        val rollup = ismRollup.toRollup(indexName)
+        rollupId = rollup.id
+        logger.info("Attempting to create a rollup job $rollupId for index $indexName")
+
+        val indexRollupRequest = IndexRollupRequest(rollup, WriteRequest.RefreshPolicy.IMMEDIATE)
+
+        try {
+            val response = withContext(Dispatchers.IO) { client.execute(IndexRollupAction.INSTANCE, indexRollupRequest).actionGet() }
```
dbbaughe, line 64: As discussed offline let's change this to use the suspendUntil helper method we have instead of actionGet

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
```diff
@@ -0,0 +1,123 @@
...
+            logger.info("Received status ${response.status.status} on trying to create rollup job $rollupId")
+
+            stepStatus = StepStatus.COMPLETED
+            info = mapOf("message" to getSuccessMessage(rollupId!!, indexName))
+        } catch (e: VersionConflictEngineException) {
+            val message = getFailedJobExistsMessage(rollupId!!, indexName)
+            logger.info(message)
+            if (rollupId == previousRunRollupId && hasPreviousRollupAttemptFailed != null && hasPreviousRollupAttemptFailed!!) {
```
dbbaughe, line 72: `hasPreviousRollupAttemptFailed != null && hasPreviousRollupAttemptFailed!!` can be changed to `hasPreviousRollupAttemptFailed == true`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
```diff
@@ -0,0 +1,123 @@
...
+        logger.info("Attempting to re-start the job $rollupId")
+        try {
+            val startRollupRequest = StartRollupRequest(rollupId!!)
+            client.execute(StartRollupAction.INSTANCE, startRollupRequest).actionGet()
+            stepStatus = StepStatus.COMPLETED
+            info = mapOf("message" to getSuccessRestartMessage(rollupId!!))
+        } catch (e: Exception) {
+            val message = getFailedToStartMessage(rollupId!!)
+            logger.error(message, e)
+            stepStatus = StepStatus.CONDITION_NOT_MET
```
dbbaughe, line 110: I think this will get this stuck in a loop if the start request fails 100% of the time for some reason, i.e. call _start, fail, set condition not met, try again next execution and repeat. Could then get very tricky trying to fix it as it never fails.

Could we set to FAILED here instead and let customer deal with transient failures by using the retry logic in the action?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
dbbaughe: What did we decide these mean again?

rollupId == previousRunRollupId is true but other is false
and
rollupId == previousRunRollupId is false but other is true

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
```diff
@@ -0,0 +1,123 @@
...
+            info = mapOf("message" to message)
+        }
+    }
+
+    companion object {
```
dbbaughe, line 115: Can we change the messaging to be consistent with others as there was work to standardize them when they were all over the place. Mainly ending with `[index=$index]` and you can probably remove the `for index` parts.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
```diff
@@ -0,0 +1,123 @@
...
+        const val name = "attempt_create_rollup"
+        fun getFailedMessage(rollupId: String, index: String) = "Failed to create the rollup job [$rollupId] for index [$index]"
+        fun getFailedJobExistsMessage(rollupId: String, index: String) = "Rollup job [$rollupId] already exists for index [$index]"
```
dbbaughe, line 118: This is the message being shown to user, in the flow above it seems like we show this even when setting to COMPLETED which means it'll move on to the next step. So can we change it to reflect what is going to happen, like Rollup job [$rollupId] already exists, skipping creation [index=$index]. Or something better that you can come up with.

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/action/RollupActionIT.kt
```diff
@@ -0,0 +1,228 @@
...
+        updateManagedIndexConfigStartTime(managedIndexConfig)
+        waitFor {
+            assertEquals(
+                    WaitForRollupCompletionStep.getJobCompletionMessage(rollupId),
+                    getExplainManagedIndexMetaData(indexName).info?.get("message")
+            )
+        }
+
+        waitFor {
+            val rollupJob = getRollup(rollupId = rollupId)
```
bowenlan-amzn, line 95: can move this waitFor into the above waitFor

bowenlan-amzn:
```
waitFor {
    val rollupJob = getRollup(rollupId = rollupId)
    assertNotNull("Rollup job doesn't have metadata set", rollupJob.metadataID)
    val rollupMetadata = getRollupMetadata(rollupJob.metadataID!!)
    assertEquals("Rollup is not finished", RollupMetadata.Status.FINISHED, rollupMetadata.status)
    assertEquals(
            WaitForRollupCompletionStep.getJobCompletionMessage(rollupId),
            getExplainManagedIndexMetaData(indexName).info?.get("message")
    )
}
```

src/main/resources/mappings/opendistro-ism-history.json
thalurur: sure, changed

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: 👍🏾

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: sure make sense - changed to FAILED

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: makes sense. changed

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/action/RollupActionIT.kt
thalurur: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/action/RollupActionConfig.kt
thalurur: kept the prefix for now, let me know if otherwise

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    override fun isIdempotent() = true
+
+    override suspend fun execute(): WaitForRollupCompletionStep {
+        val rollupJobId = managedIndexMetaData.actionMetaData?.actionProperties?.rollupId
+
+        if (rollupJobId == null) {
+            logger.error("No rollup job id passed down")
+            stepStatus = StepStatus.FAILED
+        } else {
```
dbbaughe, line 50: Is this a real use case? If not could we not just pass down the config into the step that already contains the rollup ID?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
```diff
@@ -0,0 +1,122 @@
...
+        } catch (e: VersionConflictEngineException) {
+            val message = getFailedJobExistsMessage(rollup.id, indexName)
+            logger.info(message)
+            if (rollupId == previousRunRollupId && hasPreviousRollupAttemptFailed == true) {
+                startRollupJob(rollup.id)
+            } else {
+                stepStatus = StepStatus.COMPLETED
+                info = mapOf("info" to message)
+            }
+        } catch (e: Exception) {
```
dbbaughe, line 79: Refer to some of the other steps, but you need to handle RemoteTransportExceptions and unwrap the exception.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
```diff
@@ -0,0 +1,132 @@
...
+
+                if (response.getIdsToExplain()[rollupJobId]?.metadata?.status == null) {
+                    logger.warn("Job $rollupJobId has not started yet")
+                    stepStatus = StepStatus.CONDITION_NOT_MET
+                    info = mapOf("message" to getJobProcessingMessage(rollupJobId, indexName))
+                } else {
+                    logger.info("Received metadata associated with $rollupJobId")
+                    processRollupMetadataStatus(rollupJobId, response.getIdsToExplain().getValue(rollupJobId)!!.metadata!!)
+                }
+            } catch (e: Exception) {
```
dbbaughe, line 64: Refer to some of the other steps, but you need to handle RemoteTransportExceptions and unwrap the exception.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
```diff
@@ -0,0 +1,132 @@
...
+            stepStatus = StepStatus.FAILED
+        } else {
+            val explainRollupRequest = ExplainRollupRequest(listOf(rollupJobId))
+            try {
+                val response: ExplainRollupResponse = client.suspendUntil { execute(ExplainRollupAction.INSTANCE, explainRollupRequest, it) }
+                logger.info("Received the status for jobs [${response.getIdsToExplain().keys}]")
+
+                if (response.getIdsToExplain()[rollupJobId]?.metadata?.status == null) {
+                    logger.warn("Job $rollupJobId has not started yet")
+                    stepStatus = StepStatus.CONDITION_NOT_MET
```
dbbaughe, line 58: Any chance this never exists and we end up in a cycle here with it always failing?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
```diff
@@ -0,0 +1,132 @@
...
+                    info = mapOf("message" to getJobProcessingMessage(rollupJobId, indexName))
+                } else {
+                    logger.info("Received metadata associated with $rollupJobId")
+                    processRollupMetadataStatus(rollupJobId, response.getIdsToExplain().getValue(rollupJobId)!!.metadata!!)
```
dbbaughe, line 62: Let's refactor to remove !!

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: sure, let me check

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: I am not sure of any failure where we will not set the metadata, so I am inclined to say no

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: you mean remove the log?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: I can't think of a case this will happen but programmatically it can since its an optional field. Can pass it down through the input the instead but would need to pass the whole `ISMRollup` object

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/AttemptCreateRollupJobStep.kt
thalurur: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: removed the log

dbbaughe: No I meant the .getValue(rollupJobId)!!.metadata!!
You have two uses of !! here, it's a very big code smell in Kotlin to use !! in production code and we should try to refactor the code itself to remove the usage of them.

dbbaughe: As an example above line 57 just add something like
```kotlin
val metadata = response.getIdsToExplain()[rollupJobId]?.metadata
if (metadata?.status == null) {
...
} else {
//metadata is not null here so you can pass it in without !!
}
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
```diff
@@ -0,0 +1,132 @@
...
+            RollupMetadata.Status.STOPPED -> {
+                stepStatus = StepStatus.FAILED
+                hasRollupFailed = true
+                info = mapOf("message" to getJobFailedMessage(rollupJobId, indexName), "cause" to "${rollupMetadata.failureReason}")
+            }
+        }
+    }
+
+    fun processFailure(rollupJobId: String, e: Exception) {
+        stepStatus = StepStatus.CONDITION_NOT_MET
```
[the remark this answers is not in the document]
thalurur: sure, can leave it to the customer retry policy, will update to FAILED

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
```diff
@@ -0,0 +1,132 @@
...
+            }
+            RollupMetadata.Status.FINISHED -> {
+                stepStatus = StepStatus.COMPLETED
+                info = mapOf("message" to getJobCompletionMessage(rollupJobId, indexName))
+            }
+            RollupMetadata.Status.RETRY -> {
+                stepStatus = StepStatus.CONDITION_NOT_MET
+                info = mapOf("message" to getJobProcessingMessage(rollupJobId, indexName))
+            }
+            RollupMetadata.Status.STOPPED -> {
```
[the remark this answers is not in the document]
thalurur: I can't see in the code that it does but not entirely sure about it. Do you think we should set a constant message instead - "Rollup job was stopped"?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: got it, updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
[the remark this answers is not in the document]
dbbaughe: Yeah that might be better for this enum

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
dbbaughe: Ok, can either update to use the ISMRollup or just update this to at least include a message in info object.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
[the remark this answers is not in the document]
thalurur: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/step/rollup/WaitForRollupCompletionStep.kt
thalurur: added the info message

dbbaughe (approved):

bowenlan-amzn (approved):

PR #394, opened 2021-01-27
rollup security

bowenlan-amzn:
*Issue #, if available:*

*Description of changes:*
Example Response for get and explain rollup
```
{
  "test1" : {
    "metadata_id" : null,
    "rollup_metadata" : null,
    "roles" : [
      "own_index",
      "all_access"
    ]
  },
  "test2" : {
    "metadata_id" : "02ZTiXcB-XXJjE9jiuty",
    "rollup_metadata" : {
      "rollup_id" : "test2",
      "last_updated_time" : 1612916691716,
      "status" : "finished",
      "failure_reason" : null,
      "stats" : {
        "pages_processed" : 5,
        "documents_processed" : 4675,
        "rollups_indexed" : 3627,
        "index_time_in_millis" : 741,
        "search_time_in_millis" : 323
      }
    },
    "roles" : [
      "own_index",
      "all_access"
    ]
  }
}
```

```
{
  "total_rollups" : 2,
  "rollups" : [
    {
      "_id" : "test1",
      "_seq_no" : 49,
      "_primary_term" : 8,
      "rollup" : {
        "rollup_id" : "test1",
        "enabled" : true,
        "schedule" : {
          "interval" : {
            "start_time" : 1612918410200,
            "period" : 1,
            "unit" : "Minutes"
          }
        },
        "last_updated_time" : 1612918410200,
        "enabled_time" : 1612918410200,
        "description" : "An example policy that rolls up the sample ecommerce data",
        "schema_version" : 8,
        "source_index" : "kibana_sample_data_ecommerce",
        "target_index" : "test_rollup",
        "metadata_id" : null,
        "page_size" : 1000,
        "delay" : 0,
        "continuous" : false,
...
        "roles" : [
          "own_index",
          "all_access"
        ]
      }
    },
    {
      "_id" : "test2",
      "_seq_no" : 47,
      "_primary_term" : 7,
      "rollup" : {
        "rollup_id" : "test2",
        "enabled" : false,
        "schedule" : {
          "interval" : {
            "start_time" : 1612916629413,
            "period" : 1,
            "unit" : "Minutes"
          }
        },
        "last_updated_time" : 1612916629413,
        "enabled_time" : null,
        "description" : "An example policy that rolls up the sample ecommerce data",
        "schema_version" : 8,
        "source_index" : "kibana_sample_data_ecommerce",
        "target_index" : "test_rollup",
        "metadata_id" : "02ZTiXcB-XXJjE9jiuty",
        "page_size" : 1000,
        "delay" : 0,
        "continuous" : false,
...
        "roles" : [
          "own_index",
          "all_access"
        ]
      }
    }
  ]
}
```


src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -703,12 +703,40 @@
         "continuous": {
           "type": "boolean"
         },
-        "roles": {
-          "type": "text",
-          "fields": {
-            "keyword": {
-              "type": "keyword",
-              "ignore_above": 256
+        "user": {
```
dbbaughe, line 706: Increment schema version

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
```diff
@@ -54,13 +57,19 @@ class TransportIndexRollupAction @Inject constructor(
     private val log = LogManager.getLogger(javaClass)
 
     override fun doExecute(task: Task, request: IndexRollupRequest, listener: ActionListener<IndexRollupResponse>) {
-        IndexRollupHandler(client, listener, request).start()
+        val userStr = client.threadPool().threadContext.getTransient<String>(ConfigConstants.OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT)
+        val user = resolveUser(User.parse(userStr))
+        val userName = client.threadPool().threadContext.getTransient<String>(ConfigConstants.INJECTED_USER)
```
dbbaughe, line 62: Is this pattern also used in Alerting?

bowenlan-amzn: I don't think alerting use this.
Basically, this is used because we do index rollup in runner, and we should not update user object because of that call. It will update user name to plugin.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/Rollup.kt
```diff
@@ -162,6 +163,32 @@ data class Rollup(
         return builder
     }
 
+    // hasUser indicates it is for saving rollup object with user object
```
dbbaughe, line 166: Why duplicating above method?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/Rollup.kt
```diff
@@ -162,6 +163,32 @@ data class Rollup(
...
+    // compared to get rollup API we don't show user object
+    fun toXContent(builder: XContentBuilder, params: ToXContent.Params, hasUser: Boolean): XContentBuilder {
```
dbbaughe, line 168: Move hasUser boolean into ToXContent.Params

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
dbbaughe: Where is INJECTED_USER coming from?

src/main/resources/mappings/opendistro-ism-config.json
bowenlan-amzn: increased

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/Rollup.kt
bowenlan-amzn: moved

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
bowenlan-amzn: in rollup runner index rollup method
[code link](https://github.com/opendistro-for-elasticsearch/index-management/pull/394/files?file-filters%5B%5D=.kt#diff-9d27b0547d80839891dd1315fd5491e47db004d340770bcadfcdedd131ffedbbR390)

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -209,14 +209,14 @@ object ManagedIndexRunner : ScheduledJobRunner,
...
-         * job object. `managedIndexConfig.user` will be null. Insert `all_access, AmazonES_all_access` role.
+         * job object. `managedIndexConfig.user` will be null. Insert `all_access` role.
          * 2. ISM jobs that are created when security plugin is disabled will have empty User object.
          * (`managedIndexConfig.user.name`, `managedIndexConfig.user.roles` are empty )
          * 3. ISM jobs that are created when security plugin is enabled will have an User object.
          */
         val roles = if (managedIndexConfig.user == null) {
             // fixme: discuss and remove hardcoded to settings?
-            settings.getAsList("", listOf("all_access", "AmazonES_all_access"))
+            settings.getAsList("", listOf("all_access"))
```
thalurur, line 219: minor: can you define the `listOf("all_access")` as a global variable in the package and use it everywhere instead of defining this inline

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
```diff
@@ -277,19 +279,37 @@ object RollupRunner : ScheduledJobRunner,
...
-                                val compositeRes: InternalComposite = rollupSearchResult.searchResponse.aggregations.get(updatableJob.id)
-                                metadata = metadata.incrementStats(rollupSearchResult.searchResponse, compositeRes)
-                                when (val rollupIndexResult = rollupIndexer.indexRollups(updatableJob, compositeRes)) {
-                                    is RollupIndexResult.Success -> RollupResult.Success(compositeRes, rollupIndexResult.stats)
-                                    is RollupIndexResult.Failure -> RollupResult.Failure(rollupIndexResult.message, rollupIndexResult.cause)
+                        val roles = if (job.user == null) {
+                            settings.getAsList("", listOf("all_access"))
+                        } else {
+                            job.user.roles
+                        }
```
thalurur, line 286: minor: move this to a method `Rollup.getRoles(): List<String> { .. }` and just use `job.geRoles()` here and other places fetching roles

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/explain/TransportExplainRollupAction.kt
```diff
@@ -69,60 +71,73 @@ class TransportExplainRollupAction @Inject constructor(
...
+                    try {
+                        response.hits.hits.forEach {
+                            val rollup = contentParser(it.sourceRef).parseWithType(
+                                it.id,
+                                it.seqNo,
+                                it.primaryTerm,
+                                Rollup.Companion::parse
+                            )
+                            idsToExplain[rollup.id] = ExplainRollup(metadataID = rollup.metadataID)
+                            rolesMap[rollup.id] = rollup.user?.roles
```
thalurur, line 86: Is there any reason to not show the default roles - cause that is what we are setting internally in those cases right?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
thalurur: I would prefer to define instead a variable like `shouldBackfillPermissions` or something along the line instead to make it clear the intent of doing this.

thalurur: Also I am not sure I get this entirely, why can't `putRollup` method just use the user object defined here `val user = resolveUser(User.parse(userStr))` to update the rollup object always? Shouldn't that solve the issue of backfill and also ensure the rollup stored object always have the permissions defined the last updated user?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/elasticapi/ElasticExtensions.kt
```diff
@@ -210,6 +215,7 @@ class InjectorContextElement(
 
     override fun updateThreadContext(context: CoroutineContext) {
         rolesInjectorHelper.injectRoles(roles)
+        if (user != null) rolesInjectorHelper.injectUser(user)
```
dbbaughe, line 218: Not a fan of this, this seems like it's essentially using an injected user as a toggle for the IndexRollupRequest to know whether or not it's coming from an external vs internal call. Why not use something like ConfigConstants.INTERNAL or something to be explicit since you're not actually using the injected user in any way. And if we plan to support injecting users then this needs to be removed in the future anyways.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
```diff
@@ -362,7 +382,16 @@ object RollupRunner : ScheduledJobRunner,
     private suspend fun updateRollupJob(job: Rollup, metadata: RollupMetadata): RollupJobResult {
         try {
             val req = IndexRollupRequest(rollup = job, refreshPolicy = WriteRequest.RefreshPolicy.IMMEDIATE)
-            val res: IndexRollupResponse = client.suspendUntil { execute(IndexRollupAction.INSTANCE, req, it) }
+            val roles = if (job.user == null) {
+                settings.getAsList("", listOf("all_access"))
+            } else {
+                job.user.roles
+            }
+            val res = withCloseableContext(InjectorContextElement(job.id, settings, threadPool.threadContext, roles, job.user?.name)) {
```
dbbaughe, line 390: If a job.user can be null (assuming from before this PR is introduced?) then won't null be passed in to the InjectorContextElement which then means the IndexRollupAction thinks its coming from an external request instead of internal? See above comment, let's not use "user" here for this. Prefer being explicit with an actual field for this behavior.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
dbbaughe: +1, using user is confusing

dbbaughe:
Can you also paste an example response of what the explain/get APIs look like with the roles

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/explain/TransportExplainRollupAction.kt
bowenlan-amzn: If user is null, it prob is a job from previous version. We may not need to show "all_access" roles for it here, but just call out the old job will run as before (run by "all_access" role)

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
bowenlan-amzn: refactored

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
bowenlan-amzn: good idea, refactored

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
bowenlan-amzn: changed to use a new element in thread context `INDEX_MANAGEMENT_PLUGIN_INTERNAL` to indicate if the request is from outside user or plugin runner.
if the value of this element is true, then we will not update user object.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/RollupRunner.kt
bowenlan-amzn: introduce explicit flag for this `INDEX_MANAGEMENT_PLUGIN_INTERNAL`

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/elasticapi/ElasticExtensions.kt
```diff
@@ -210,6 +225,7 @@ class InjectorContextElement(
 
     override fun updateThreadContext(context: CoroutineContext) {
         rolesInjectorHelper.injectRoles(roles)
+        threadContext.putTransient(INDEX_MANAGEMENT_PLUGIN_INTERNAL, internalReq)
```
dbbaughe, line 228: Need to remove this when suspending; also prob need to do a quick sanity check to make sure it isn't already there otherwise you'll get corruption exceptions

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/rollup/model/XContentTests.kt
```diff
@@ -117,6 +117,7 @@ class XContentTests : ESTestCase() {
     fun `test rollup parsing with type`() {
         val rollup = randomRollup()
         val rollupString = rollup.toJsonString()
+        println("rollup string: $rollupString")
```
dbbaughe, line 120: Remove

dbbaughe (approved):

thalurur (approved):

PR #402, opened 2021-02-10
Introducing transform data model

thalurur:
*Issue #, if available:*
N/A

*Description of changes:*
Introducing a transform data model and tests

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/Transform.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/Transform.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
 package com.amazon.opendistroforelasticsearch.indexmanagement.transform.model
 
 import com.amazon.opendistroforelasticsearch.indexmanagement.elasticapi.instant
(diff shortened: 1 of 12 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/Transform.kt
```diff
@@ -0,0 +1,308 @@
...
+import org.elasticsearch.index.query.AbstractQueryBuilder
+import org.elasticsearch.index.query.MatchAllQueryBuilder
+import org.elasticsearch.index.query.QueryBuilder
+import org.elasticsearch.index.seqno.SequenceNumbers
+import org.elasticsearch.search.aggregations.AggregatorFactories
+import java.io.IOException
+import java.time.Instant
+
+
+data class Transform (
```
qreshi, line 37: Given the large overlap of attributes between `Rollup` and `Transform`, should we create some base class for materialized views and have both of these data classes extend it to cover their differences without too much duplication? Or are they intentionally being kept separate?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/Transform.kt
```diff
@@ -0,0 +1,317 @@
...
+    Writeable {
+
+    init {
+        aggregations.aggregatorFactories.forEach {
+            require(supportedAggregations.contains(it.type)) { "Unsupported aggregation [${it.type}]" }
+        }
+        when (jobSchedule) {
+            is CronSchedule -> {}
+            is IntervalSchedule -> {
+                require(jobSchedule.interval >= MINIMUM_JOB_INTERVAL) { "Transform job schedule interval must be greater than 0" }
```
dbbaughe, line 65: Hmm is this not already a requirement in job scheduler? If not I feel like this should be added there instead of here.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/TransformMetadata.kt
```diff
@@ -0,0 +1,215 @@
...
+                lastUpdatedAt = requireNotNull(lastUpdatedAt) { "Last updated time must not be null" },
+                status = requireNotNull(status) { "Status must not be null" },
+                failureReason = failureReason,
+                stats = requireNotNull(stats) { "Stats must not be null" }
+            )
+        }
+    }
+}
+
+data class TransformStats(
```
dbbaughe, line 140: nit: Could probably split this out to own file

dbbaughe: Also is this exact copy of the RollupStats by any chance? If so perhaps we can make something like CompositeStats data class that these can extend or something?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/Transform.kt
thalurur: sure, will add in the job scheduler if not present.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/model/TransformMetadata.kt
thalurur: Sure, will move it. Its not an exact copy, but has the same intent. RollupStats named the field `rollupsIndexed` instead of `documentsIndexed`.
I was wondering about it too, if its okay to rename the field in rollup document (not sure if any issues with changing schema for existing rollup jobs) then I can have have a common data class for both transforms and rollups

dbbaughe (approved):

PR #417, opened 2021-03-11
Transform api get

lobdelle:
*Issue #, if available:*
Get Transform API

*Description of changes:*
Added Get Transform API action, transport action, request, response, Rest action, unit tests, and integ tests

All unit and integ tests pass and build succeeds.

    indicated in the file; or

    in the file; or

    it.


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/action/get/TransportGetTransformAction.kt
```diff
@@ -0,0 +1,67 @@
...
+        val getRequest = GetRequest(INDEX_MANAGEMENT_INDEX, request.id)
+            .fetchSourceContext(request.srcContext).preference(request.preference)
+        client.get(getRequest, object : ActionListener<GetResponse> {
+            override fun onResponse(response: GetResponse) {
+                if (!response.isExists) {
+                    return listener.onFailure(ElasticsearchStatusException("Transform not found", RestStatus.NOT_FOUND))
+                }
+
+                var transform: Transform? = null
+                if (!response.isSourceEmpty) {
```
thalurur, line 41: Looks like this happens if the index we are searching from has the `_source` disabled which shouldn't happen here since the system index where we store these docs have the source enabled. I think it would be better to throw an Internal exception instead if source is empty instead of returning empty transform object.

https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/action/get/TransportGetTransformsAction.kt
```diff
@@ -0,0 +1,85 @@
...
+class TransportGetTransformsAction @Inject constructor(
+    transportService: TransportService,
+    val client: Client,
+    actionFilters: ActionFilters,
+    val xContentRegistry: NamedXContentRegistry
+) : HandledTransportAction<GetTransformsRequest, GetTransformsResponse> (
+    GetTransformsAction.NAME, transportService, actionFilters, ::GetTransformsRequest
+) {
+
+    override fun doExecute(task: Task, request: GetTransformsRequest, listener: ActionListener<GetTransformsResponse>) {
```
thalurur, line 41: This probably is same as `rollup` logic can you instead create an util method to hold this and reuse this in both rollups and here?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/action/get/TransportGetTransformsAction.kt
```diff
@@ -0,0 +1,85 @@
...
+                    }
+                }
+            }
+
+            override fun onFailure(e: Exception) = listener.onFailure(e)
+        })
+    }
+
+    private fun contentParser(bytesReference: BytesReference): XContentParser {
+        return XContentHelper.createParser(NamedXContentRegistry(SearchModule(Settings.EMPTY, false, emptyList()).namedXContents),
```
thalurur, line 82: Can you check if the `xContentRegistry` passed in to the class can be used here?

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/resthandler/RestGetTransformActionIT.kt
```diff
@@ -0,0 +1,108 @@
...
+            getTransform(ESTestCase.randomAlphaOfLength(20))
+            fail("Expected response exception")
+        } catch (e: ResponseException) {
+            assertEquals(RestStatus.NOT_FOUND, e.response.restStatus())
+        }
+    }
+
+    @Throws(Exception::class)
+    fun `test getting all transforms`() {
+        val transforms = randomList(1, 15) { createTransform(randomTransform()) }
```
thalurur, line 44: may be add a TODO here to do delete before this test to clear up the existing transforms

src/test/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/TransformRestTestCase.kt
```diff
@@ -89,5 +103,38 @@ abstract class TransformRestTestCase : IndexManagementRestTestCase() {
...
+                _PRIMARY_TERM -> primaryTerm = parser.longValue()
+                Transform.TRANSFORM_TYPE -> transform = Transform.parse(parser, id, seqNo, primaryTerm)
+            }
+        }
+        return transform
+    }
+
     protected fun Transform.toHttpEntity(): HttpEntity = StringEntity(toJsonString(), APPLICATION_JSON)
+
+    protected fun newParser(response: Response): XContentParser {
```
thalurur, line 136: optional:

You can re-use the parser method in the ESTestCase this class is extending by just overriding the xContentRegistry method in ESTestCase to return the correct registry


```
// Where ever you are using newParsers(response) you could call instead
createParser(XContentType.JSON.xContent(), response.entity.content)

// and instead of newParser method you could have following override
override fun xContentRegistry(): NamedXContentRegistry {
        return NamedXContentRegistry(SearchModule(Settings.EMPTY, false, emptyList()).namedXContents)
    }

```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/action/get/TransportGetTransformsAction.kt
lobdelle: Wrote common logic and applied it to Transform class. Haven't applied logic to Rollup class yet.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/util/GetUtils.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.elasticsearch.common.bytes.BytesReference
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.index.query.BoolQueryBuilder
+import org.elasticsearch.index.query.ExistsQueryBuilder
+import org.elasticsearch.index.query.WildcardQueryBuilder
+import org.elasticsearch.rest.RestStatus
+import org.elasticsearch.search.builder.SearchSourceBuilder
+import org.elasticsearch.search.sort.SortOrder
+
+fun executeJobsSearch(
```
thalurur, line 25: Thanks for the refactor. 

Also may be rename the class to `ScheduledJobUtils`

I think the method signature is a little too big, may be we can do something along these lines

```
        fun getJobs (
            client: Client,
            searchSourceBuilder: SearchSourceBuilder,
            listener: ActionListener<ActionResponse>,
            scheduledJobType: String
        ) {
            val searchRequest = SearchRequest(IndexManagementPlugin.INDEX_MANAGEMENT_INDEX).source(searchSourceBuilder)
            client.search(searchRequest, object : ActionListener<SearchResponse> {
                override fun onResponse(response: SearchResponse) {
                    val totalJobs = response.hits.totalHits?.value ?: 0

                    if (response.shardFailures.isNotEmpty()) {
                        val failure = response.shardFailures.reduce { s1, s2 -> if (s1.status().status > s2.status().status) s1 else s2 }
                        listener.onFailure(ElasticsearchStatusException("Get $scheduledJobType failed on some shards", failure.status(), failure.cause))
                    } else {
                        try {
                            val jobs = response.hits.hits.map {
                                contentParser(it.sourceRef).parseWithType(it.id, it.seqNo, it.primaryTerm, getParser(scheduledJobType))
                            }
                            listener.onResponse(populateResponse(scheduledJobType, jobs, RestStatus.OK, totalJobs.toInt()))
                        } catch (e: Exception) {
                            listener.onFailure(
                                ElasticsearchStatusException("Failed to parse $scheduledJobType", RestStatus.INTERNAL_SERVER_ERROR, ExceptionsHelper.unwrapCause(e))
                            )
                        }
                    }
                }

                override fun onFailure(e: Exception) = listener.onFailure(e)
            })
        }

        private fun populateResponse(
            jobType: String,
            jobs: List<Any>,
            status: RestStatus,
            totalJobs: Int
        ): ActionResponse {
            return when (jobType) {
                Rollup.ROLLUP_TYPE -> GetRollupsResponse(jobs as List<Rollup>, totalJobs, status)
                else -> {
                    // throw Exception
                }
            }
        }

        private fun getParser(jobType: String): (XContentParser, String, Long, Long) -> Any {
            return when (jobType) {
                Transform.TRANSFORM_TYPE -> Transform.Companion::parse
                Rollup.ROLLUP_TYPE -> Rollup.Companion::parse
                else -> {
                    // throw Exception
                }
            }
        }

        private fun contentParser(bytesReference: BytesReference): XContentParser {
          return XContentHelper.createParser(NamedXContentRegistry.EMPTY,
              LoggingDeprecationHandler.INSTANCE, bytesReference, XContentType.JSON)
       }

```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/transform/action/get/TransportGetTransformAction.kt
thalurur: Looks like we are doing a ``HEAD`` REST call which doesn't return the actual transform but just the id and primary/seq no. For now ignore my comment. 

We should not throw the exception but rather just return empty in those cases

lobdelle: Updated to include this logic and also properly handle HEAD calls.

thalurur (approved):

Issue #429, opened 2021-04-02
[BUG] Problem with applying templates for more than 10 ISM policys


mousemaxx:
**Describe the bug**
"index_patterns" for ISM template for eleventh or more are not applied
Probably is the issue, the internal search for ISM templates (policies) is not specifying the size so it’s defaulting to only getting 10 of them.

**Other plugins installed**
ODFE 1.13.1

**To Reproduce**
The same policies are consistently created:
“policy_1” … “policy_11” with patterns “test_1- *” … “test_11- *”
the text does not matter, for example:
`{

“policy”: {
“policy_id”: “policy_1”,
“description”: “delete index after 7 days”,
“default_state”: “norm”,
“states”: [
{“name”: “norm”,“actions”: ,“transitions”: [{“state_name”: “del”,“conditions”: {“min_index_age”: “7d”}}]},
{“name”: “del”,“actions”: [{“delete”: {}}],“transitions”: }
],
“ism_template”: {
“index_patterns”: [“test_1-*”],
“priority”: 50
}
}
}`

after creating each policy, an index is created
**PUT test_X-1**
For the first 10, the policy is assigned, for the 11, no. Moreover, if, after creating 11 policies, you change any of the previously created ones, they also cease to be assigned
**GET _opendistro/_ism/explain/test_***
{
“test_1-1”: {
“index.opendistro.index_state_management.policy_id”: “**policy_1**”,
“index”: “test_1-1”,
“index_uuid”: “X1n6d3_YSLymYWHJVqTBow”,
“policy_id”: “policy_1”
},
“test_2-1”: {
“index.opendistro.index_state_management.policy_id”: “**policy_2**”,
“index”: “test_2-1”,
“index_uuid”: “h9olWdoPTYygyqjdlYVx2Q”,
“policy_id”: “policy_2”
},
“test_3-1”: {
“index.opendistro.index_state_management.policy_id”: “**policy_3**”,
“index”: “test_3-1”,
“index_uuid”: “t_Ejz_kMSna69o1iC3-o0A”,
“policy_id”: “policy_3”
},
“test_4-1”: {
“index.opendistro.index_state_management.policy_id”: “**policy_4**”,
“index”: “test_4-1”,
“index_uuid”: “q2WmtiqjQLysYZ-2miSXdQ”,
“policy_id”: “policy_4”
},
“test_5-1”: {
“index.opendistro.index_state_management.policy_id”: “**policy_5**”,
[...]

PR #432, opened 2021-04-06
auto migrate ISM template

bowenlan-amzn:
*Issue #, if available:*
#429 

*Description of changes:*
This PR is mainly for helping previous user who use legacy index template to auto manage index by ISM plugin. We help auto migrate the legacy index template to ISM templates.

### Main logic:
1. Find out all templates with `opendistro.index_state_management.policy_id` setting
2. Group these templates by its template priority
    1. Within each priority bucket, increase priority based on the order. For example, if we have template [t1, t5, t10] in one bucket, because their priority are all 0. Then based on their order, we add 2 to t1's priority, add 1 to t5's pirority, add 0 to t10's priority, also record the largest increment of this bucket, which is 2. These operations are maintain the order in the same bucket.
    2. Between buckets with different priority, to maintain their relation, higher order bucket's all templates' priority need to be increased by the summation of all lower order buckets' largest increments. For example, order 0 bucket with largest increment 2, then for order 1 bucket, all templates in it, their priority needs to be increased by 2.
3. Transform `templateName -> [pattern, processed_order, policyID]` to `policyID -> listOf(ISMTemplate(pattern, processed_order, last_updated_time))`

this whole migration will be executed by this master node

Introduce cluster setting `opendistro.index_state_management.template_migration.enabled`
If you set this setting to negative value -1/-2, it will cancel the scheduled job.

If you set this setting to `>= 0` value, it will start the migration:
- `= 0`: use the master node election time cached in Coordinator as `last_updated_time`
- `> 0`: use this settings value as `last_updated_time`
You can check the setting value later:
- `< 0`: -1 means ends successfully; -2 means migration has been done 10 times but not succeeded, need human intervention

### Enhance ISM Template:
Now one policy can have multiple ISM templates, they will be saved in list in `policy.ism_templates` field.
The restriction for ISM template is the same: templates with overlapping index pattern and having same order are not allowed, please choose different non-overlapping pattern or different order.

---

    indicated in the file; or

    in the file; or

    it.


thalurur:
Can you check the failed tests?

src/test/resources/mappings/cached-opendistro-ism-config.json
```diff
@@ -443,7 +443,8 @@
             }
           }
         },
-        "ism_template": {
+        "ism_templates": {
```
dbbaughe, line 446: I think you have a few choices here.

What you current have done is a breaking change as there will be people in ODFE that are using "ism_template". We should try do avoid this.

Another option is to duplicate them, and in the parser you read in both and add them together in the data model.

Another option is to just not rename ism_template -> ism_templates.

And then there is an alias field type too, but that would need to be looked in to for how it behaves and if it could have any issues.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/Policy.kt
```diff
@@ -155,7 +162,22 @@ data class Policy(
...
+                            when (xcp.currentToken()) {
+                                Token.START_ARRAY -> {
+                                    while (xcp.nextToken() != Token.END_ARRAY) {
+                                        ismTemplates.add(ISMTemplate.parse(xcp))
+                                    }
+                                }
+                                Token.START_OBJECT -> {
+                                    ismTemplates.add(ISMTemplate.parse(xcp))
+                                }
+                                else -> ensureExpectedToken(Token.START_ARRAY, xcp.currentToken(), xcp)
```
dbbaughe, line 177: Hmm what's the point of this else branch?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -98,30 +144,360 @@ fun validateFormat(indexPatterns: List<String>): ElasticsearchException? {
...
+    private val policiesFailedToUpdate = mutableMapOf<policyID, BulkItemResponse.Failure>()
+    private lateinit var lastUpdatedTime: Instant
+
+    @Volatile private var retryPolicy =
+        BackoffPolicy.constantBackoff(TimeValue.timeValueMillis(50), 3)
+
+    suspend fun doMigration(timeStamp: Instant) {
+        logger.info("Doing ISM template ${runTimeCounter++} time.")
+        if (runTimeCounter > 10) {
+            stopMigration(-2)
```
dbbaughe, line 231: Return after stop

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -98,30 +144,360 @@ fun validateFormat(indexPatterns: List<String>): ElasticsearchException? {
...
+        updateISMPolicies()
+
+        if (policiesToUpdate.isEmpty()) {
+            stopMigration(-1)
+        }
+
+        cleanCache()
+    }
+
+    private fun stopMigration(successFlag: Long) {
```
dbbaughe, line 252: Reset stop counter

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -98,30 +144,360 @@ fun validateFormat(indexPatterns: List<String>): ElasticsearchException? {
...
+        if (order < 0) return negOrderToPositive[order] ?: 0
+        return order + (negOrderToPositive.size)
+    }
+
+    private fun bucketizeV1TemplatesByOrder() {
+        clusterService.state().metadata.templates.forEach {
+            if (v1TemplatesWithPolicyID[it.key] != null) {
+                val priority = normalizePriority(it.value.order)
+                // cache the non-negative priority
+                v1TemplatesWithPolicyID[it.key] = v1TemplatesWithPolicyID[it.key]!!.copy(order = priority)
```
dbbaughe, line 326:
```kotlin
val templateCache = v1TemplatesWithPolicyID[it.key]
if (templateCache != null) {
    ...
    v1TemplatesWithPolicyID[it.key] = templateCache.copy(order = priority)
    ...
}
```

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -98,30 +144,360 @@ fun validateFormat(indexPatterns: List<String>): ElasticsearchException? {
...
+
+                val bucket = v1orderToTemplatesName[priority]
+                if (bucket == null) {
+                    v1orderToTemplatesName[priority] = mutableListOf(it.key)
+                } else {
+                    // add later one to start of the list
+                    v1orderToTemplatesName[priority]!!.add(0, it.key)
```
dbbaughe, line 333: bucket

src/test/resources/mappings/cached-opendistro-ism-config.json
bowenlan-amzn: still using ism_template, but now internally we treat it as a List, not a single object

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/Policy.kt
bowenlan-amzn: just for sanity, when user input not array, not object, showing an exception telling them you should input array

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
bowenlan-amzn: added

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
bowenlan-amzn: added

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
bowenlan-amzn: resolved

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
bowenlan-amzn: resolved

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -98,30 +139,368 @@ fun validateFormat(indexPatterns: List<String>): ElasticsearchException? {
...
+            requestsToRetry = retryItems.map { bulkReq.requests()[it] as UpdateRequest }
+
+            if (requestsToRetry.isNotEmpty()) {
+                val retryCause = failedResponses.first { it.status() == RestStatus.TOO_MANY_REQUESTS }.failure.cause
+                throw ExceptionsHelper.convertToElastic(retryCause)
+            }
+        }
+    }
+
+    private fun cleanCache() {
```
dbbaughe, line 488: I think in a few places we don't catch all exceptions or we throw exceptions during the migration process. This would mean the cache was not cleared and these in memory structures are filled with data. If that happens and then the migration runs again, will we get incorrect results? Should we clear the cache at the start of every migration attempt?

bowenlan-amzn: Good point, we should clear cache at the beginning of every run.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -98,30 +139,368 @@ fun validateFormat(indexPatterns: List<String>): ElasticsearchException? {
     return null
 }
 
+fun List<ISMTemplate>.findSelfConflictingTemplates(): Pair<List<String>, List<String>>? {
+    val priorityList = this.map { it.priority }
+    priorityList.forEach { priority ->
+        // same priority
+        val indexPatternsList = this.filter { it.priority == priority }.map { it.indexPatterns }
```
qreshi, line 146: If I'm not mistaken, if we're trying to check the indexPatterns of any templates with the same priority, it might be more efficient to create a Map of `priority -> List<ISMTemplate>` first so that way you can iterate through the priorities and then only check the indexPatterns when the list size is greater than 1 for a particular priority key in the Map.

This will help avoid doing `this.filter` each time just to get matching priorities.

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
```diff
@@ -45,6 +46,18 @@ class ManagedIndexSettings {
             Setting.Property.Dynamic
         )
 
+        // 0: enabled, use onMaster time as ISM template last_updated_time
+        // -1: migration ended successfully
+        // -2: migration ended unsuccessfully
+        // >0: use this setting (epoch millis) as ISM template last_updated_time
+        val TEMPLATE_MIGRATION_CONTROL: Setting<Long> = Setting.longSetting(
```
qreshi, line 53: From what I've seen this setting is doing multiple things:
1. Determining whether or not template migration is enabled
2. Configuring migration behavior (like what template last_updated_time is used)
3. Propagating migration success/failure status

Since this setting is user configurable as well, is this kind of usage conventional or are we overloading the use of it?

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/model/Policy.kt
```diff
@@ -43,7 +43,7 @@ data class Policy(
     val errorNotification: ErrorNotification?,
     val defaultState: String,
     val states: List<State>,
-    val ismTemplate: ISMTemplate? = null
+    val ismTemplate: List<ISMTemplate>? = null
```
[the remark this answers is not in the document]
bowenlan-amzn: keep it still ismTemplate to be consistent with `ism_template` in the config index mapping

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -38,28 +78,29 @@ private val log = LogManager.getLogger("ISMTemplateService")
  * @return policyID
  */
 @Suppress("ReturnCount")
-fun Map<String, ISMTemplate>.findMatchingPolicy(indexMetadata: IndexMetadata): String? {
+fun Map<String, List<ISMTemplate>>.findMatchingPolicy(indexName: String, indexCreationDate: Long, isHiddenIndex: Boolean): String? {
```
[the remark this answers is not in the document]
bowenlan-amzn: updated

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
bowenlan-amzn: The idea is:
User can set it to `0` value to do migration; set it to `-1` to stop migration
After starting the migration, the process is quick, like minutes after, they can check if it goes well from the value `-1/-2`

If the value set to a positive value, it will be used as the take effect time for ISM template
for example: 1 -> `Thursday, January 1, 1970 12:00:01 AM`
ISM template will auto manage all index created after this time, if the index matches the index pattern in the ISM template.

I think this should be a fairly straight forward experience from user's point of view

src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
bowenlan-amzn: sure, changed

qreshi (approved):

dbbaughe (approved):

qreshi (approved):

PR #436, opened 2021-04-08
Fixes ISM template matching

dbbaughe:
ISM template matching was only getting the first 10 policies

*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/index-management/issues/429

*Description of changes:*

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
+++ b/src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
@@ -315,12 +315,11 @@ class ManagedIndexCoordinator(
     suspend fun getISMTemplates(): Map<String, ISMTemplate> {
         val searchRequest = SearchRequest()
             .source(
-                SearchSourceBuilder().query(
-                    QueryBuilders.existsQuery(ISM_TEMPLATE_FIELD)
-                )
+                SearchSourceBuilder()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/amazon/opendistroforelasticsearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -315,9 +315,9 @@ class ManagedIndexCoordinator(
     suspend fun getISMTemplates(): Map<String, ISMTemplate> {
         val searchRequest = SearchRequest()
             .source(
-                SearchSourceBuilder().query(
-                    QueryBuilders.existsQuery(ISM_TEMPLATE_FIELD)
-                )
+                SearchSourceBuilder()
+                .query(QueryBuilders.existsQuery(ISM_TEMPLATE_FIELD))
+                .size(MAX_HITS) // Assuming there are not more than 10k policies for now
```
thalurur, line 320: Do you think have a cluster setting with default as MAX_HITS instead? in case users want to exceed this or its an overkill for now

dbbaughe: I think overkill... for one I don't know of anyone really having over 10,000 policies and if a user sets it too low then they could run into the issue that was reported for this. Instead we'll probably have to scroll through and load all the policies, but that's a bit more work and I wanted to get out a quick fix for the user that is affected.

dbbaughe:
Bowen has fixed in his other PR, will close this one.

